PR #150, opened 2026-05-03
Multi-select cards in search with bulk add to collection

mgzwarrior:
## Summary
- Adds a Select mode to the card search page so users can pick many cards at once.
- "Select all matching" picks every card matching the active filters across all pages.
- New backend endpoint `POST /api/collection/bulk-add` adds them in one round-trip, reusing the existing dedupe rules (matching card_id+variant+lang+condition+purchase_price increments quantity instead of creating a duplicate row).

## Changes
- **Backend** (`backend/api/collection.py`, `backend/schemas.py`): bulk-add endpoint returns `{ added, updated, failed, errors }`. Error handling is per-item — one bad card doesn't fail the whole batch.
- **Frontend** (`frontend/src/pages/CardSearch.jsx`):
  - Header "Select" toggle (only visible when a search is active).
  - In select mode, clicking a card toggles a red checkmark instead of opening the detail modal.
  - Sticky toolbar: count, **Select page** / **Select all matching (N)** / **Clear** / **Add to collection**.
  - Per-card `lang` is preserved when `langFilter='all'` is mixing EN+DE results.
- **i18n** (en/de/zh): new strings.
- **API client** (`frontend/src/api/client.js`): `bulkAddToCollection(items)` helper.

## Notes for reviewer
- Bulk add uses default settings (quantity=1, condition=NM, no variant, no purchase_price) — users can edit individual entries afterward in the collection. A future iteration could add a settings modal before commit if needed.
- "Select all matching" refetches with `page_size = total_count` to materialize IDs on the client. For very large result sets this is a single, large request rather than many paginated ones; acceptable for this app's typical search volume.
- I couldn't run the JS build locally (no `node_modules` in either tree). Python files parse cleanly via AST and the JSX was visually reviewed; please run `npm run build` to confirm before merging.

## Test plan
- [ ] Run a search, click **Select**, click several cards — checkmarks appear and toolbar count updates.
- [ ] Click **Select page**, paginate, click **Select page** again — selections persist across pages.
- [ ] Click **Select all matching (N)** — all matching cards across pages are selected.
- [ ] With a mixed-language search (`langFilter='all'`), bulk-add and verify EN/DE cards land with correct `lang` in the collection.
- [ ] Bulk add cards already in the collection — quantities increment, no duplicate rows.
- [ ] Click **Done** to exit select mode — selection clears, normal click reopens the detail modal.


Git-Romer:
Thanks for the PR! I tested the updated branch on my instance and the bulk select/add flow seems to work.

I also pushed two small maintainer fixes into this PR:
- bumped the app version to v1.7
- adjusted bulk-add transaction handling so one failed card should not roll back the whole batch

Frontend build also passes locally.

I’ll keep testing a bit, but this looks good to me so far.

PR #159, opened 2026-05-15
Add downloadable debug logs

Git-Romer:
## Summary
- add admin-only debug mode setting
- write backend debug logs to a rotating file when enabled
- add admin-only debug log download endpoint
- add Settings UI toggle and download button
- restore debug logging state on backend startup

## Validation
- python3 -m compileall backend
- npm --prefix frontend run build
- verified debug logging writes an entry to the log file when enabled

## Notes
- Debug logs are backend-side only for now.
- Existing backend logger calls are captured while debug mode is enabled.
- A request logging middleware also writes debug entries, so navigating the app produces log output.


Git-Romer:
Addressed the practical Copilot review findings in 5187b78:

* defer debug-mode side effects until after the settings transaction commits
* keep container/stdout handlers at INFO while the debug file captures DEBUG entries
* restore previous logger and handler levels when debug mode is disabled
* add no-store headers to the debug log download endpoint
* make the frontend Blob download cleanup safer by revoking the object URL asynchronously

The multi-worker note is valid generally, but the current Docker command runs a single Uvicorn worker, so the in-process toggle matches the current deployment model.

PR #195, opened 2026-05-22
Implement issue 193 QoL improvements

### LINKED ISSUE #193: A few QoL Improvements maybe?
**Collection**

- Sort by card ID (POR 001, 002, ...)
- “Clone” the card with different properties (Normal/Holo/Reverse Holo)
- Maybe add to binders directly from the card menu via a dropdown of existing binders?

**Set Overview**

- I like the Set Overview because it lets me sort new cards so quickly. However, it does happen every now and then that you click the wrong button and add a card you didn’t mean to add.
It would be great to have the option to remove a card from the collection right from the Set Overview.
Maybe there could be a menu in one of the card corners where you could directly select “Remove from collection”/trash can icon, and next to it the option to add another version of the card—similar to “Clone” in the collection, but with different properties (a plus icon similar to when a card hasn’t been added yet; clicking it doesn’t add the card directly but opens a menu similar to the collection menu)

**Binder**

- I like this feature precisely because it gives me an overview of which cards I have as Toploaders or in the Master Collection. But adding cards is a bit of a hassle. It would be nice if I could search using the same criteria as in the collection itself.
So, by card ID with filters for sets, etc., or perhaps being able to select from the Collection’s card menu: “This is in a binder --> dropdown.”
- It would be super cool to have different icons for the binders/cases.

**Card Search**

- It would be great if the card search could show right away whether I already have the card in my collection. At the very latest when I open the card in the search results and see the “Add” menu, a cross-check against my collection would be really nice.

That's what we noticed. I wonder if anyone else has any ideas on the best way to go about this?

Git-Romer:
## Summary
- Add safe Set Overview actions so card clicks open an add/remove menu instead of immediately adding cards.
- Add collection card-ID sorting, clone/add-another-copy controls, and collection-item-to-binder dropdowns.
- Add exact collection-item binder entries plus Pokémon GIF binder icons.
- Show owned status/quantity in card search results and the card add modal.
- Bump app/API version to 1.12.

## Notes
- Binder entries can now reference a specific `collection` item, preserving variant/condition for collection binders.
- Set Overview remove decrements quantity by one when quantity is above 1, otherwise removes the exact collection item.

## Validation
- `python3 -m py_compile backend/api/binders.py backend/api/cards.py backend/api/sets.py backend/models.py backend/schemas.py backend/database.py backend/main.py`
- `git diff --check`

Closes #193


Git-Romer:
@copilot can you check this PR like you did with other PR's?

PR #244, opened 2026-05-30
Feat: adds arrowkey navigation

djinyus:
Adds arrowkeys navigation (left, right) in CardSearch.jsx

PR #245, opened 2026-05-30
Update UI after adding a card to collection

djinyus:
When a card is added to the collection from the search results, the counter isn't updated in the search results because the search query cache isn't being invalidated.

This should fix it.


Git-Romer:
Thanks a lot for the contribution, I really appreciate you taking the time to work on this.

I am going to close this PR, because it mixes unrelated changes. The PR is about updating the card search counter after adding a card, but it also includes the arrow-key navigation commit.

This is similar to the general issue in your other PR #244, where unrelated changes and a SQL dump were included. Please make sure future PRs are based on a fresh clone or a fresh branch from the latest `main`, and only include the specific change the PR is about.

If you want to contribute this fix, please open a new clean PR that only contains the counter update change.

Even better, feel free to open an issue describing the bug or feature request and what behavior you would like to see. I would be very happy to implement it cleanly, and I would still be glad to include you in the contributors list for suggesting and helping shape the feature.

Thanks again, I do appreciate the help.

PR #249, opened 2026-05-31
added arrow key pagination in CardSearch

djinyus:
## Summary

- Add left and right arrow key pagination to Card Search.
- Ignore arrow pagination while inputs are focused or modifier keys are held.
- Disable arrow pagination while Card Search overlays are open, including card details, filters, custom card creation, and scanner.
- Bump app version to `1.21.4`.

## Validation

- `node scripts/check-version.mjs`
- `git diff --check`
- `PYTHONPATH=backend python -m unittest discover backend/tests`
- Smoke test with local mock API and headless Chromium: verified ArrowRight moves Card Search from page 1 to page 2, and ArrowLeft does not change the page while the card modal or filter sheet is open.


Git-Romer:
Thanks for cleaning this up and making the PR focused.

I found one small follow-up while testing: the arrow key pagination should not run while overlays are open, because it can change the search page behind a card modal, filter sheet, custom card modal, or scanner.

I will handle that cleanup and the version bump on my side, then merge once checks pass.

Thanks again for the contribution.

PR #254, opened 2026-06-01
<select> styling in Settings.jsx updated to match 

djinyus:
Updated the <select> styling in the settings.jsx to match with the app design. When opening the <select> in the settings page, the styling was not on par and switched to a white background for example.


Git-Romer:
Thanks for the styling improvement here. I added a follow-up pass to make the PR complete and keep the release metadata consistent.

## What changed in this PR now

- Updated Settings dropdowns to use the shared dark `.select` styling instead of one-off inline styling.
- Restored a visible dropdown chevron for custom-styled selects after `appearance: none` hid the native arrow.
- Centralized dark `<option>` styling in the shared `.select option` CSS.
- Switched the Settings user-role dropdown to the shared `.select` class as well.
- Added a `select-no-arrow` helper for the disabled “No collection binders” placeholder so it keeps the select-like box styling without showing a misleading dropdown arrow.
- Bumped the app version to `v1.22.3` in `VERSION`, README, `frontend/package.json`, and `frontend/package-lock.json`.

## Validation

- `node scripts/check-version.mjs`
- `git diff --check`
- Fresh reviewer pass completed; the only cleanup note was fixed before pushing.
- GitHub checks are passing and the PR is mergeable (`CLEAN`).

PR #284, opened 2026-07-14
Resolve catalogue display language per user, not globally (#283)

### LINKED ISSUE #283: Per-user language is ignored by /api/sets and /api/cards (English UI shows German card data)
## Summary

In multi-user mode, a user's per-user `language` setting controls the UI strings but **not** the catalogue data. `/api/sets` and `/api/cards` resolve the display language from the **global** `settings` row instead of the requesting user's `user_settings`, so a user set to English gets an English interface populated with German card and set names (or vice versa).

## Steps to reproduce

1. Fresh install of v1.22.13 (global `settings.language` defaults to `de`).
2. As the admin, set language to English in Settings (this writes `user_settings.language = 'en'` for that user; the global row stays `de`).
3. Browse sets/cards.

## Expected

Card and set names are shown in the language the logged-in user selected.

## Actual

UI chrome is English, but the catalogue data comes back German — e.g. Base Set is displayed as **"Grundset"** (`base1_de`) rather than **"Base Set"** (`base1_en`).

A second user with no explicit override (e.g. a freshly created `trainer`) inherits the global `de` and gets an entirely German UI, which is surprising on an otherwise-English install.

## Root cause

`backend/api/sets.py` and `backend/api/cards.py` both define:

```python
def _get_language(db: Session) -> str:
    """Get display language from settings."""
    row = db.query(Setting).filter(Setting.key == "language").first()
    return row.value if row else "de"
```

That queries the **global** `Setting` table. It never consults `UserSetting` for the current user. It is then used to pick which language rows to serve, e.g. `backend/api/sets.py`:

```python
display_lang = lang_filter if lang_filter != "all" else _get_language(db)
```

Since v42 migrated `language` into the per-user settings (`per_user_keys` in `database.py` includes `"language"`), the two now disagree, and the catalogue endpoints read the wrong one.

---

## Update: the global row is unreachable, and there are three defaults, not one

Digging further, the mismatch is not just that the two paths disagree — **the global row the catalogue endpoints read can never be changed through the app at all.**

**1. `settings.language` is write-once.** `language` is in `PER_USER_KEYS`, not `ADMIN_ONLY_KEYS`. Both write paths in `api/settings.py` — `update_settings` (`PUT /`) and `set_setting` (`POST /{key}`) — send any key that isn't in `ADMIN_ONLY_KEYS` to `user_settings`. So after `init_db()` seeds the global row, **no UI action ever writes it again.** It is frozen at the seeded value (`de`) for the life of the install, and `_get_language()` in `sets.py`/`cards.py` is pinned to that value no matter what any user picks.

Conversely, `_get_user_settings()` only loads global rows whose key is in `ADMIN_ONLY_KEYS`, so it never reads `settings.language` either. The global row's **only** consumer is the two catalogue endpoints. That's why the SQL workaround below is currently the *only* way to affect them.

**2. The default lives in three places, which disagree.** Anyone fixing this should know there are two separate `DEFAULT_SETTINGS` dicts with different jobs:

| Location | Role |
|---|---|
| `database.py:33` | seeds the global `settings` table on init (INSERT-if-not-exists) |
| `api/settings.py:44` | `setdefault` fallback in the read path, only applies when the key is **absent** |
| `sets.py` / `cards.py` | hardcoded `else "de"` in each `_get_language()` |

**3. Trap: changing only `api/settings.py` looks like it works, then silently regresses on the first restart.** I hit this while patching locally. Because `main.py`'s lifespan calls `init_db()` **before** `bootstrap_admin()`, on a genuinely first boot no admin exists yet, so the v42 migration finds no admin and is skipped — the admin ends up with no `user_settings.language` row and correctly falls through to the `setdefault`. On the **second** boot the admin now exists, the migration copies the seeded global `de` into `user_settings.language`, and that explicit row shadows the default permanently.

Simulating three consecutive boots against a clean DB, with only `api/settings.py` changed to `en`:

```
boot 1: settings.language='de'  user_settings.language=None  -> GET /settings = 'en'  | card/set names in 'de'
boot 2: settings.language='de'  user_settings.language='de'  -> GET /settings = 'de'  | card/set names in 'de'
boot 3: settings.language='de'  user_settings.language='de'  -> GET /settings = 'de'  | card/set names in 'de'
```

English on install, German forever after the first `docker compose restart`. Note the last column: even on boot 1, while the UI is English, the catalogue is already German — this issue, reproduced from a clean database.

## Suggested fix

Resolve the display language per-user, falling back to the global setting and then to the default — i.e. `_get_language(db, user_id=current_user.id)`, consistent with how `get_gemini_key(db, user_id=...)` already handles a per-user key in `recognize.py`.

Given the above, a complete fix should also align the seed default in `database.py` with the one in `api/settings.py` and the hardcoded `else` fallbacks in `sets.py`/`cards.py`, so that a fresh install is internally consistent regardless of boot count. (Changing them all to `en` and re-running the three-boot simulation gives `en` in every column on every boot; the 117-test backend suite still passes.)

Worth considering whether the vestigial global `settings.language` row should be dropped altogether now that nothing can write it — it exists only to be misread.

## Workaround

Set the global row to match, so both paths agree. This is currently the only way to move it, since the API never writes it:

```sql
UPDATE settings SET value = 'en' WHERE key = 'language';
UPDATE user_settings SET value = 'en' WHERE key = 'language';
```

## Environment

- PokéCollector v1.22.13, self-hosted via `docker compose`, `multi_user_mode = true`, 2 users (admin + trainer)

hiddenbanana:
Fixes #283.

## The bug

`/api/sets` resolved its display language from the **global** `settings` row — but that row is unwritable from the app. `language` is in `PER_USER_KEYS`, so both `update_settings` (`PUT /`) and `set_setting` (`POST /{key}`) route it to `user_settings`. The global row is seeded once by `init_db()` and then frozen forever, so the catalogue stayed pinned to the seed value no matter what any user picked.

That matters more than it first looks, because `_refresh_sets()` fetches **only** the resolved language (`languages = [normalized_display_lang]`). On a fresh install the catalogue is therefore populated in that one language — so an English user ends up with a German-only catalogue and sees **"Grundset"** instead of **"Base Set"**.

## The fix

New `services/display_language.py`:

```python
def get_display_language(db: Session, user_id: int | None = None) -> str:
    """Resolve the catalogue display language for a user."""
```

It prefers the user's own `user_settings` row, falls back to the global `settings` row, then to the default — the shape suggested in the issue, and consistent with how `get_gemini_key(db, user_id=...)` already handles a per-user key. `api/sets.py` now calls it with `current_user.id`.

## Two things found along the way

**`api/cards.py` defined its own `_get_language` that was never called.** The issue text (mine) claimed both endpoints used it; only `sets.py` did. Removed the dead function and its now-unused `Setting` import.

**The seeded default in `database.py` disagreed with the one in `api/settings.py`** (`"de"` vs `"en"`). This is a live trap rather than cosmetic: `main.py`'s lifespan calls `init_db()` **before** `bootstrap_admin()`, so on a genuinely first boot no admin exists and the v42 migration is skipped — the admin falls through to `api/settings.py`'s default. On the *second* boot the admin exists, the migration copies the seeded global value into `user_settings.language`, and that explicit row shadows the default permanently. A mismatch here means the admin's language silently changes on the first `docker compose restart`. Aligned both.

Simulating three consecutive boots against a clean DB, before and after:

```
before:  boot 1 -> GET /settings = 'en' | catalogue 'de'
         boot 2 -> GET /settings = 'de' | catalogue 'de'     <- flipped on restart
         boot 3 -> GET /settings = 'de' | catalogue 'de'

after:   boot 1 -> GET /settings = 'en' | catalogue 'en'
         boot 2 -> GET /settings = 'en' | catalogue 'en'
         boot 3 -> GET /settings = 'en' | catalogue 'en'
```

## Testing

New `tests/test_display_language.py` covers per-user preference, two users resolving independently, global fallback, default fallback, and the no-user case. Written test-first; verified failing (`ModuleNotFoundError`) before implementing.

Full backend suite: **122 passing** (117 existing + 5 new). Frontend builds clean.

## Note on the second commit

Commit 2 (`Default the UI to English…`) is **independently droppable** if you'd rather keep this PR strictly to the bug. It aligns the frontend default with the backend and mirrors the chosen language into `localStorage`, so the first paint — and an unauthenticated or failed settings fetch — uses the language the user actually picked instead of flashing the default. Happy to split it out if you prefer.

One thing worth your call: the global `settings.language` row is now vestigial. Nothing can write it, and after this change nothing meaningful reads it except as a fallback. It may be worth dropping entirely — it exists mainly to be misread. Left in place here as a conservative fallback.


Git-Romer:
Maintainer update: I pushed commit 800ceab onto this PR instead of continuing with the duplicate PR #288. Thanks again to hiddenbanana for the original report and PR.

Added on top of the original PR:
- /api/cards now also resolves omitted lang from the requesting user language.
- /api/sets now distinguishes omitted lang from explicit lang=all.
- Added regression tests for set/card omitted language and explicit lang=all behavior.
- Frontend set queries now use language-aware cache keys and avoid accidental full-catalogue searches from language hydration.
- Version metadata is bumped and checked at 1.23.2.

Validation:
- Fresh reviewer pass found the omitted-vs-all /api/sets blocker, now fixed.
- git diff --check
- Python compile check for changed backend files
- Backend suite in temporary venv: 129 tests OK
- npm run build passes with the existing large chunk warning

Temporary test environment was removed after validation; no local server is running. Duplicate PR #288 has been closed.

Git-Romer:
Final maintainer update after merging current main into this branch:

- PR #284 is now mergeable.
- Duplicate PR #288 remains closed.
- Final validation on the merged branch:
  - git diff --check
  - Python compile check for changed backend files
  - Backend suite in temporary venv: 144 tests OK
  - npm run build passes with the existing large chunk warning
- Temporary test environment was removed after validation; no local server is running.

PR #288, opened 2026-07-15
Fix per-user catalogue language

### LINKED ISSUE #283: Per-user language is ignored by /api/sets and /api/cards (English UI shows German card data)

Git-Romer:
## Summary
- Fixes #283
- Resolve catalogue language from each users per-user language setting for sets/cards when no explicit language filter is provided
- Keep explicit lang=all behavior for all visible catalogue languages
- Align fresh-install language defaults to English and bump version metadata to 1.23.2
- Make frontend set-data cache keys language-aware and avoid accidental full catalogue card searches

## Validation
- Fresh reviewer pass completed, follow-up reviewer completed after cleanup
- git diff --check
- Python compile check for changed backend files
- Backend test suite in temporary venv: 137 tests OK
- npm run build passes with the existing large chunk warning

Temporary test environment was removed after validation; no local server is running.


Git-Romer:
Closing this duplicate in favor of hiddenbanana’s original PR #284, which now contains the completed fixes and validation updates.

PR #291, opened 2026-07-15
Add experimental Pokemon Center queue alerts

Git-Romer:
## Summary
- add an experimental Pokemon Center queue monitor with states: normal, queue, bot_protection, unknown, and error
- alert opted-in users through their per-user Telegram config only when the monitor transitions into queue state
- add verbose server logging and an admin status/manual-check panel so this branch can be tested during a real Pokemon Center drop
- store detector evidence in `pokemon_center_queue_status` for later troubleshooting
- add a one-hour notification cooldown and bot-protection false-positive guards
- bump version metadata to v1.23.5

## Important experimental note
The exact Pokemon Center queue HTML is not currently verifiable outside a live drop. This branch intentionally keeps extra logging and admin-visible evidence so it can be deployed temporarily and validated during the next real queue event. Once the detector is confirmed, the debug surface can be reduced before finalizing the feature.

## Safety boundary
This only detects and notifies. It does not join queues, bypass waiting rooms, solve challenges, reserve spots, or automate checkout.

## Validation
- fresh reviewer pass found blockers; fixed them
- follow-up reviewer pass found no blockers
- git diff --check
- Python compile check for touched backend files
- lightweight detector tests: `PYTHONPATH=backend python -m unittest backend.tests.test_pokemon_center_queue`
- backend full suite in temp venv: 158 tests OK
- frontend `npm run build`
- live detector check from this host correctly classified Pokemon Center as `bot_protection`, not `queue`
- disposable Postgres migration check confirmed `pokemon_center_queue_status` is created
- temp venv and Postgres container removed; no local server is running


Git-Romer:
Closing and deleting the experimental queue-alert branch as requested.

PR #294, opened 2026-07-19
feat: skip full sync when one is already running

### LINKED ISSUE #293: Full sync has no cross-process guard: concurrent runs crash with duplicate card primary key
## Summary

`perform_full_sync` has no guard against two full syncs running at the same time. The scheduler's periodic full sync and an admin-triggered manual sync can overlap, and because both insert the same rows into the shared `cards` table, one run aborts with a duplicate-primary-key `IntegrityError`. The aborted run's transaction is left un-rolled-back, so its `sync_log` row is never finalised and stays stuck at `status = "running"`.

## Steps to reproduce

1. Start a full sync (e.g. the scheduled one, or `POST /api/sync/`).
2. While it is still running, start a second full sync from a different path — e.g. the scheduler fires while a manual run is in progress, or `perform_full_sync` is invoked in a separate process/worker.
3. Let both proceed past the set/card catalogue stage.

## Expected

Only one full sync runs at a time. A second trigger is skipped while one is already running, and the catalogue sync completes.

## Actual

The two runs race on the shared `cards` table and one aborts:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "cards_pkey"
DETAIL:  Key (id)=(sm9-1_pt) already exists.
```

followed by a cascade of `PendingRollbackError`. Because the failing transaction is never rolled back, that run's `sync_log` row is never updated and remains `status = "running"` indefinitely, so later runs have to treat it as an orphaned row.

## Root cause

The only concurrency guard is the in-process module global `_sync_running` in `backend/api/sync.py`:

```python
_sync_running = False
...
def trigger_sync(...):
    global _sync_running
    if _sync_running:
        return {"message": "Sync already running", "status": "running"}
```

This guard has two gaps:

**1. It only covers the manual API path.** The scheduler calls `perform_full_sync` directly, without checking or setting `_sync_running` (`backend/services/scheduler.py`):

```python
def run_full_sync():
    ...
    from services.sync_service import perform_full_sync
    perform_full_sync(db)
```

**2. It is per-process.** A module global does nothing across multiple workers/processes, so any second process that reaches `perform_full_sync` proceeds anyway.

`perform_full_sync` itself performs no check — it goes straight to creating a `running` `SyncLog` and syncing (`backend/services/sync_service.py`):

```python
def perform_full_sync(db: Session) -> dict:
    """Perform a full sync cycle: sets + cards + prices."""
    log = SyncLog(started_at=datetime.datetime.utcnow(), status="running", sync_type="full")
    ...
```

So a scheduled run and a manual run (or two processes) race each other on the `cards` primary key.

## Proposed fix

Add a DB-level guard at the top of `perform_full_sync` that checks the shared `sync_log` table for a recently-started `running` full sync and returns early if one exists. Using shared DB state rather than an in-process flag covers every caller — scheduler, API, and direct/manual invocation — and is process-safe. A staleness window (e.g. 60 minutes) ensures an orphaned `running` row left behind by a crashed run cannot block future syncs forever.

I have a tested implementation and am happy to open a PR if that's welcome.

## Related

The same `cards_pkey` duplicate error can also occur in a **single** full-sync run (no concurrency involved): when a set's native card count is below its total, the cross-language fallback re-clones a card that was just added natively in the same transaction. Because the session uses `autoflush=False`, the fallback's "already exists" check doesn't see the pending native rows and stages a second insert of the same id. That is a separate root cause from the concurrency issue above and can be filed/fixed separately.

hiddenbanana:
Fixes #293.

## Summary

Adds a DB-level guard so two full syncs cannot run concurrently and race on card primary keys. `perform_full_sync` now checks the shared `sync_log` table for a recently-started running full sync and returns early if one is already in progress.

Using shared DB state (rather than the existing in-process `_sync_running` flag, which only guards the manual API path) means this covers **every** caller — the scheduler, the API, and direct/manual invocation — and is process-safe.

A `running` full-sync row older than 60 minutes is treated as orphaned (e.g. left behind by a crash or container restart), so a stale row can never block future syncs forever.

## Testing

New `backend/tests/test_full_sync_guard.py` (6 tests, `unittest`):

- no rows → not in progress
- recent running full → blocks
- stale running full (>60m) → does **not** block
- running *price* sync → does not block a full sync
- finished full → does not block
- `perform_full_sync` short-circuits to `{"status": "skipped"}` without creating a new log row or doing any network work when a full sync is already running

`python -m unittest tests.test_full_sync_guard` → 6/6 passing.

### DIFF
--- a/backend/services/sync_service.py
+++ b/backend/services/sync_service.py
@@ -502,8 +502,32 @@ def check_custom_card_matches(db: Session):
             logger.warning(f"Failed to check API match for custom card {card.id}: {e}")
 
 
+# A full sync takes ~20 min. A "running" full-sync row older than this is
+# treated as orphaned (left by a crash or container restart) so a stale row can
+# never block future syncs forever.
+FULL_SYNC_STALE_AFTER = datetime.timedelta(minutes=60)
+
+
+def _full_sync_in_progress(db: Session) -> bool:
+    """True if another full sync is actively running.
+
+    Guards against two overlapping full syncs racing and colliding on card
+    primary keys. Works across processes and callers (scheduler, API, manual)
+    because it checks the shared sync_log table rather than an in-process flag.
+    """
+    cutoff = datetime.datetime.utcnow() - FULL_SYNC_STALE_AFTER
+    return db.query(SyncLog.id).filter(
+        SyncLog.sync_type == "full",
+        SyncLog.status == "running",
+        SyncLog.started_at >= cutoff,
+    ).first() is not None
+
+
 def perform_full_sync(db: Session) -> dict:
     """Perform a full sync cycle: sets + cards + prices."""
+    if _full_sync_in_progress(db):
+        logger.info("Full sync skipped: another full sync is already running")
+        return {"cards_updated": 0, "sets_updated": 0, "status": "skipped"}
     log = SyncLog(started_at=datetime.datetime.utcnow(), status="running", sync_type="full")
     db.add(log)
     db.commit()
--- a/backend/services/sync_service.py
+++ b/backend/services/sync_service.py
@@ -1,9 +1,10 @@
 import logging
 import datetime
 import math
+from contextlib import contextmanager
 from typing import Any, Mapping
 from sqlalchemy.orm import Session, load_only
-from sqlalchemy import func, or_
+from sqlalchemy import func, or_, text
 from models import Card, Set, CollectionItem, WishlistItem, BinderCard, PriceHistory, SyncLog, PortfolioSnapshot, CustomCardMatch, ProductPurchase, User, UserSetting
 from services import pokemon_api, telegram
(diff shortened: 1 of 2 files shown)


Git-Romer:
Thanks for the PR and for digging into the root cause here. The issue is valid, but the original read-before-insert `sync_log` guard still had a small race window.

I pushed a follow-up commit (`ce3a675`) to make the guard atomic and safer:

* Full sync now uses a PostgreSQL advisory lock, held on a dedicated DB connection for the whole run
* The stale `sync_log` check remains as a second safety layer
* Failed full syncs now roll back before marking the `SyncLog` as `error`
* Lock release failure invalidates the DB connection so a held advisory lock cannot be returned to the pool
* Full-sync guard tests were expanded to cover lock acquire, release, busy-lock, and failure logging cases

PR #295, opened 2026-07-19
Unify owned variants and wishlist indicators across card tiles

hiddenbanana:
## Summary

Unifies collector state indicators across card tiles. This PR now contains both the original variant-pill work and the card-state indicator work consolidated from #298.

Cards can show:

- Owned print variants in canonical order: Normal, Holo, Reverse Holo, First Edition
- Total quantity per owned variant when more than one copy is owned
- A generic owned indicator when an endpoint only has aggregate ownership
- A wishlist heart independently of ownership

The shared indicators are used in Set Detail, Card Search, and the reusable CardItem component used by Pokédex and other card grids.

## Backend

- Adds batched, collector-scoped card-state summaries
- Aggregates duplicate collection rows by variant without exposing another user's state
- Excludes zero or negative legacy quantities from ownership
- Adds ownership variants and wishlist state to Card Search and Set checklist payloads
- Rejects new collection quantities outside 1–999

## Frontend

- Adds a reusable `CardStateIndicators` component
- Keeps variant ordering and normalization in one helper
- Refreshes all cached card-tile views after collection, wishlist, trade, migration, product-sale, scan, CSV import, and bulk-add mutations
- Keeps detailed variants ahead of generic ownership fallbacks
- Preserves unknown/custom variants instead of hiding them

## Dependency and release maintenance

- Merged current `main`
- Preserves Axios 1.18
- Updates Vitest to 4.1.10, removing the prior high/critical development dependency findings
- Regenerates `package-lock.json`
- Bumps the app version to 1.24.0

## Validation

- Frontend: 18 tests passed
- Backend: all 199 tests passed in the Python 3.11 production image
- Frontend production build passed locally and in the Node 20 Docker build
- Backend production Docker image built successfully
- Version consistency check passed
- `npm audit`: 0 high, 0 critical; two pre-existing moderate React Router advisories remain and require a separate major-version migration


engelben:
ah just saw this, fits nicely with the other pills I am proposing in #298. Will try to double check tomorrow if we can combine this cleanly.

hiddenbanana:
be nice to combine. like to see what you do with it :) 

still feel the pills arent as pretty as they could be 

image

Git-Romer:
@hiddenbanana @engelben Thanks to both of you. I honestly love seeing this.

It is already amazing that people are contributing to PokéCollector, but seeing contributors also talk to each other and coordinate ideas/features is even better. This is exactly the kind of community direction I hoped the project could grow into, and I really hope it keeps getting bigger and reaches more collectors over time.

That being said, it looks like #295, #296, and #298 are still being discussed, tested, or refined. I’ll wait for now and won’t merge these until you consider your respective work ready.

Once you are done, please leave a quick note in the relevant PR saying it is approved from your side / ready for final review. After that I’ll do my own testing pass and merge when everything looks good.

engelben:
I think #295 and #298 could fit together really nicely if we turn them into one shared card-state component instead of having separate badge systems.

My rough idea would be:

* top-left: owned variants as compact icons, with `×N` only when there’s more than one copy
* top-right: a small wishlist heart
* when variant details are available, they replace the generic owned check and total quantity badge
* when we only have `owned: true`, we fall back to the simple green check
* fallback-language/data badges stay at the bottom

Implementation-wise, #295 already has most of the useful foundation:

* `getOwnedVariants()` and the variant metadata/order
* the new variant indicator component
* tests for aggregation and unknown variants
* Set Detail already has `owned_items`, so it can show the full variant detail immediately

I could then adapt #298 so `CardItem` uses the same shared component in Card Search and Pokédex instead of its local `CardStatusBadges`.

Something along these lines:

```jsx
<CardStateIndicators
  card={card}
  compact={compact}
/>
```

The component could normalize whatever data is available:

```js
owned_variants
owned_items
owned / owned_quantity
wishlisted
```

So Set Detail gets the detailed variant icons, while Search/Pokédex can initially fall back to the green check until their API response also includes an `owned_variants` summary.

For Set Detail I’d also remove the existing top-left quantity badge and probably the tiny owned check once variant icons are shown, otherwise we end up with quantity + checkmark + variant pills all saying roughly the same thing.

That would keep both contributions, avoid duplicate UI, and give us one consistent system across Set Detail, Card Search and Pokédex.

What do you reckon?

hiddenbanana:
[engelben](https://github.com/engelben) i love your plan. completely think this is a great way to move forward :)

engelben:
> [engelben](https://github.com/engelben) i love your plan. completely think this is a great way to move forward :)

Great :) I’ll prepare the unified implementation as a PR against your #295 branch, so #295 stays the single consolidated route to main. Could you rebase the branch onto the latest upstream main first? Once that’s done I’ll branch from it, bring over the relevant parts of #298 and add the shared component/API handling.

engelben:
PR in your fork created @hiddenbanana - have a look. Once it gets merged there we can flag it here for merge.

hiddenbanana:
PR Merged 👍

engelben:
@Git-Romer ready for your review. We consolidated changes from #295 and #298 in the latest commits to this branch.

Updated list of changes:

## Summary

This builds on the variant-indicator work from #295 and combines it with the owned/wishlist card-state work from #298.

The result is one shared indicator system across:

- Set Detail
- Card Search
- Pokédex card results
- regular and compact card tiles

Owned cards now show their actual owned variants and per-variant quantities where that data is available. Wishlist state is shown independently with a compact heart indicator.

## What changed

### Shared card-state indicators

Introduces a shared `CardStateIndicators` component instead of maintaining separate badge implementations per view.

The component supports:

- owned variants and quantities;
- generic owned fallback when detailed variant data is unavailable;
- wishlist state;
- simultaneous owned and wishlisted state;
- regular and compact card layouts.

Detailed variant indicators replace the generic owned check whenever variant information is available. The component also supports future `wishlist_count` payloads while continuing to use the existing `wishlisted` boolean contract.

### Set Detail

- shows owned variants and their quantities;
- shows wishlist state;
- refreshes the exact Set Detail checklist after collection and wishlist mutations;
- removes duplicate ownership indicators;
- retains existing owned-card styling and fallback badges;
- leaves hover behavior above the non-interactive indicator layer.

### Card Search

- replaces the previous generic total-owned badge with shared variant-specific indicators;
- shows wishlist state;
- refreshes indicators after collection or wishlist changes;
- moves the bulk-selection checkbox to the bottom-left of the image so it does not overlap the top-left variant indicators or top-right wishlist heart.

### Pokédex

- continues to render species printing cards through the shared `CardItem` path;
- receives the same detailed `owned_variants` contract through the Card Search API query;
- preserves the existing owned/wishlist refresh behavior.

### Backend card-state summaries

Relevant tile responses now provide consistent, current-user-scoped card state:

```json
{
  "owned": true,
  "owned_quantity": 3,
  "owned_variants": [
    {
      "variant": "Normal",
      "quantity": 2
    },
    {
      "variant": "Reverse Holo",
      "quantity": 1
    }
  ],
  "wishlisted": true
}
```

## UI consistency discussion

There is still a broader difference between the card views:

* Pokédex and Collection render cards inside a framed tile with additional card data below the image.
* Set Detail and Card Search render the card image more directly and rely more heavily on image overlays.

This PR does not attempt to redesign or consolidate those tile layouts. The shared indicators work in the current layouts, but a future UI pass could establish a common card-tile structure across all views.

## Possible follow-ups

These are intentionally outside the scope of this PR:

### Collection grouping

Add an optional grouped Collection view, likely grouping by card printing and language and showing the owned variants and quantities through the same indicators.

This would make the indicators useful in Collection without rendering every owned copy as a separate tile.

### Available-variant filtering

The Pokédex card modal already filters the variant dropdown according to the variants actually available for a card.

The same behavior should be extracted into shared logic and applied to:

* Set Detail;
* Card Search;
* Collection add/edit flows.

Backend validation should also reject impossible card/variant combinations.

### Wishlist management in the card modal

The card modal currently provides richer management for owned copies than for wishlist state.

A follow-up could add a dedicated wishlist section that supports:

* viewing current wishlist membership;
* removing the card from the wishlist;
* adjusting wanted quantity;
* later selecting and managing multiple named wishlists.

This would also provide a natural foundation for a future multiple-wishlist feature similar to Cardmarket.

PR #298, opened 2026-07-20
Show collection and wishlist status on card tiles

engelben:
## Summary

Adds compact ownership and wishlist indicators to shared card tiles.

## Changes

- shows an Owned badge when the current user owns the card
- includes the owned quantity when greater than one
- shows a Wishlist badge when the card is on the current user's wishlist
- supports both normal and compact card layouts
- reuses the ownership and wishlist information already returned by card search APIs

## Motivation

Card search and species views currently require opening a card to determine whether it is already owned or wishlisted. The badges make this state visible directly in search and browsing views.

## Scope

This is a frontend-only presentation change. It does not alter collection or wishlist data.

## Validation

- frontend production build passes
- verified cards with no status show no additional overlay
- verified Owned, Wishlist, and combined states
- verified compact and regular card layouts

image

### DIFF
--- a/frontend/src/components/CardItem.jsx
+++ b/frontend/src/components/CardItem.jsx
@@ -390,6 +390,46 @@ export function CustomCardModal({ onClose, onCreated, sets: setsProp = [], autoA
   )
 }
 
+function CardStatusBadges({ card, compact = false, t }) {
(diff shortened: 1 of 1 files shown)


engelben:
Closing this PR as it was consolidated in #295

PR #312, opened 2026-07-27
Restore Pokédex and Sets list position after detail navigation

engelben:
## Summary

Restore list state and scroll position when returning from Pokémon and set detail pages.

This updates the following navigation flows:

- `/pokedex` → `/pokedex/:dexId`
- `/sets` → `/sets/:setId`

## Changes

- Save the list’s current scroll offset, route, query string, and originating item anchor in `sessionStorage` before opening a detail page.
- Restore the saved position only when returning to the matching history entry.
- Wait until asynchronously loaded list content has rendered before restoring.
- Use the originating Pokémon or set as an anchor fallback when the saved pixel offset can no longer be restored exactly.
- Clear the restoration request after it has been consumed.
- Add source-list navigation state to Pokémon and set detail links.
- Make detail-page Back buttons traverse history when the detail was opened from its corresponding list.
- Keep safe `/pokedex` and `/sets` fallbacks for direct detail-page visits and refreshed detail pages.
- Preserve Pokédex search and ownership filters while returning from a detail page.
- Preserve the originating Pokédex overview while navigating through multiple Previous/Next Pokémon.
  - Browser Back and Forward continue to traverse individual Pokémon detail entries normally.
  - The detail page’s Back button returns directly to the originating Pokédex overview.
- Ensure newly pushed detail pages start at the top without overriding pending list restoration during Back navigation.

## Implementation

A shared `useListScrollRestoration` hook owns:

- session-storage keys and validation;
- scroll-position and list-state persistence;
- route/history-entry matching;
- asynchronous restoration readiness;
- exact-offset restoration;
- item-anchor fallback;
- consumed-state cleanup;
- shared detail Back behavior.

Saved restoration requests are scoped by both list name and React Router location key. This prevents a stale Pokédex or Sets position from being applied during a fresh navigation to the list.

Restoration runs after the list reports that it is ready and after two animation frames, allowing asynchronously fetched content to render before applying the saved offset.

### Pokédex

1. Open `/pokedex`.
2. Apply a search, generation, or ownership filter.
3. Scroll down and open a Pokémon.
4. Navigate through several Pokémon using Previous and Next.
5. Click the detail page’s Back button.
6. Confirm the original overview, filters, and scroll position are restored.
7. Repeat with browser Back and Forward.

### Sets

1. Open `/sets`.
2. Apply filters or sorting.
3. Scroll down and open a set.
4. Wait for the set checklist to load.
5. Click the detail page’s Back button.
6. Confirm the original Sets position is restored.
7. Repeat with browser Back and Forward.

### Direct detail entry

1. Open a Pokémon or set detail URL directly.
2. Click the page’s Back button.
3. Confirm it navigates safely to `/pokedex` or `/sets`.


Git-Romer:
Thanks, @engelben, for the contribution. I kept your core sessionStorage and anchor-based restoration approach and added several follow-up improvements:

• Preserved restoration state across repeated browser Back/Forward cycles instead of consuming it after the first return.
• Restored the complete Sets filter and sorting state associated with the original history entry.
• Saved pending Sets preferences immediately before opening a set.
• Tracked Previous/Next detail navigation depth so the page’s Back button returns directly to the originating overview.
• Kept browser Back/Forward navigation working normally between individual Pokémon detail pages.
• Made the Pokédex generation URL the source of truth, ensuring browser history changes update both the selected generation and displayed content.
• Added validation for malformed saved state and regression tests for these navigation cases.
• Integrated the current main branch and updated the release version.

I also completed final desktop, mobile, automated, and build validation. Everything relevant to this PR is now working as intended. Thanks again!

PR #321, opened 2026-08-07
fix: add cooldown to metadata enrichment so uncacheable cards stop being re-fetched

### LINKED ISSUE #318: Metadata enrichment re-fetches uncacheable cards on every search, forever (no cooldown)
## Summary

`card_needs_metadata_enrichment` has no cooldown, so cards whose metadata TCGdex genuinely does not provide are re-fetched from the TCGdex API on **every search request that returns them**, forever. On my install 10,664 of 31,603 cards (33.7%) are permanently in this state, and any search page containing one pays ~1s per such card on every single request.

## Reproduction

Search for a term whose first page contains affected cards, twice in a row:

```
pika    1960 ms   total_count=204
pika    1820 ms   total_count=204   ← no improvement
```

Compare with a query whose page converges normally:

```
eevee    927 ms   total_count=107
eevee     37 ms   total_count=107   ← self-healed after one request
```

Counting outbound requests during a repeat `pika` search confirms it is re-fetching, not cached:

```
tcgdex calls during pika search: 2
GET https://api.tcgdex.net/v2/en/cards/smp-SM170
GET https://api.tcgdex.net/v2/en/cards/smp-SM190
```

The rows, after the fetch:

```
id            rarity  supertype  subtypes  updated_at
smp-SM170_en  Promo   Pokemon    null      19:34:38   ← rewritten every search
smp-SM190_en  Promo   Pokemon    null      19:34:39   ← rewritten every search
```

These are promo cards with no `subtypes` in TCGdex. The fetch succeeds, returns nothing for that field, the row is written back still missing it, and it qualifies again on the next search.

## Cause

`services/card_metadata.py` — `subtypes` is a key field:

```python
_KEY_METADATA_FIELDS = ("rarity", "supertype", "subtypes")

def card_needs_metadata_enrichment(card: Card) -> bool:
    ...
    if any(not _has_value(getattr(card, field, None)) for field in _KEY_METADATA_FIELDS):
        return True
```

`enrich_cards_metadata` looks like it guards against repeat attempts — it bumps `updated_at` on both the `missing` and `failed` paths:

```python
else:
    result["missing"] += 1
    card.updated_at = datetime.datetime.utcnow()
    db.add(card)
```

But `card_needs_metadata_enrichment` only tests field presence and never reads `updated_at`, so that bump has no effect. There is no backoff anywhere in the path.

Because `_enrich_search_page_metadata` (`api/cards.py:148`) runs this on the search response path with `limit=METADATA_ENRICHMENT_PER_SEARCH_PAGE = 20`, each affected card costs one **synchronous** TCGdex round trip inside the request. `pokemon_api.get_card` also opens a fresh `httpx.Client` per call, so there is no connection reuse:

```python
def get_card(card_id, lang="en"):
    with httpx.Client(timeout=30.0) as client:
        response = client.get(f"{base_url}/cards/{card_id}")
```

The same wall shows up in the Pokédex backfill's own tally — `attempted 22140, updated 22139, missing 1` — that one row can never be satisfied either.

## Scale on this install

```
cards (non-custom)                      31603
subtypes = JSON null                    10664   (33.7%)
  of which lang=ja                       8159
  of which lang=en                       2505
```

Japanese cards are disproportionately affected. Worst observed single query: 4663 ms for `jigglypuff` (41 results).

## Suggested fix

Add a cooldown to `card_needs_metadata_enrichment` — skip a card whose `updated_at` is within some window (a few days), so a card TCGdex has nothing more for is retried occasionally rather than on every request. The `updated_at` bump already in `enrich_cards_metadata` is presumably intended for exactly this and just isn't consulted.

I have not opened a PR because the right window length is a product decision. Two adjacent things worth considering separately:

- Moving per-search enrichment off the request path entirely (background queue), so a cold card never blocks a user's search.
- Reusing a module-level `httpx.Client` in `pokemon_api` for connection pooling — currently every enrichment pays a fresh TCP + TLS handshake.

## Environment

- v1.27.4 (backend carries one unrelated local patch to `api/recognize.py`; nothing in the enrichment path)
- PostgreSQL 18, Docker Compose
- `TCGDEX_SYNC_LANGUAGES=en,ja`
- Measurements taken with the full sync and the one-time Pokédex backfill both complete and zero background TCGdex traffic (verified: 0 requests in 15s idle).

lamiskin:
## Why

Cards whose full TCGdex detail is unavailable or permanently incomplete were fetched again whenever they appeared in search. Reusing the general `updated_at` field for a cooldown was unsafe because catalogue, image, and price updates also change it.

Closes #318.

## What changed

- Added an indexed `last_metadata_enrichment_attempt_at` field and upgrade migration.
- Records every metadata attempt independently, including successful, missing, and failed requests.
- Applies a seven-day retry cooldown without allowing unrelated card updates to postpone retries.
- Returns search results immediately and performs eligible enrichment afterward through a FastAPI background task with its own database session.
- Uses an atomic conditional database claim so concurrent searches and full syncs cannot fetch the same incomplete card twice.
- Keeps full-sync work bounded and fair: never-attempted cards first, then the oldest eligible attempts.
- Preserves explicit Pokédex backfill and manual `--refresh` behavior while preventing stale workers from claiming the same card.
- Adds Lamiskin to `CONTRIBUTORS.csv` and bumps the application to `1.33.1`.

Images remain on the set-catalogue synchronization path and do not use this per-card metadata queue.

## Validation

- 390 backend tests pass (3 skipped).
- 159 frontend tests pass.
- Translation-key check passes.
- Production frontend build passes.
- Version consistency and `git diff --check` pass.
- Fresh isolated reviewer pass completed; all blocking findings were fixed and the re-review is ready.


Git-Romer:
Hi @lamiskin, and thanks again for reporting and investigating this.

Cards with unavailable or permanently incomplete TCGdex metadata will no longer be fetched during every search. Attempts are now tracked separately, retried after a seven-day cooldown, and processed in the background so searches remain responsive. Unrelated price, image, or catalogue updates will not affect this cooldown.

PR #322, opened 2026-08-07
Show the owner's own photo for cards

lamiskin:
To let owners see a photo of the actual card they have, not just the catalogue scan.

## Why

Roughly one card in fourteen has no TCGdex image — trainer kits, Japanese printings, the newest energy subsets — and the collection currently renders an anonymous card back for every one of them. A visual collection where a slice of the cards are indistinguishable rectangles is doing much less than it looks.

## What

The owner can attach a photo of the actual card. Once attached, it's what the collection shows — the photo always wins over the catalogue scan, not just when one is missing. A catalogue scan is a reference image of the printing; the owner's photo is evidence of the actual physical card they have (condition, centering, whatever makes their copy theirs), and that's worth seeing even when TCGdex also has a scan. It's shown everywhere that item's picture appears — Collection (grid, list, detail), Dashboard, HomeScreen, Binders (grid, detail modal, add-from-collection picker, print-optimization preview, equivalent-prints), and Analytics' duplicates page — always badged with a camera marker so it's never confused with a catalogue scan.

Attaching a photo works two ways:
- **Manually**, from the collection item's detail dialog (Manage tab) — upload or remove at any time, for any card.
- **Automatically from the scanner.** Confirming a scan match attaches the photo you just took as the item's own photo, for both the single-photo and batch-queue flows. Best-effort (a failed attach never blocks the add) and never overwrites a photo an existing (grouped) item already has.

Design notes, since these were deliberate rather than incidental:

- **Photo hangs off the collection row, not the card.** `cards` is a catalogue shared by every user of the instance, and `/api/images` is mounted without authentication (it gates on set/language visibility, not on a user) — a photo stored there would be served to anyone who could reach the instance. A photo of a card the owner has is data that belongs to the collector, not the catalogue.
- **Separate table, own endpoint.** The collection listing only carries a `has_scan_photo` flag; the bytes come from their own endpoint scoped to the owner, so a listing never drags whole JPEGs along to answer a yes/no question. Deleting a collection item cascades to its photo.
- **Re-encoded on upload**, stripping EXIF/XMP/ICC (GPS coordinates, camera serial, timestamps — none of which describe a Pokémon card, and all of which would otherwise leave the instance on export/share). Orientation is applied before the tag is discarded, so phone photos don't land on their side. Re-encoding also bounds size (a phone photo vs. ~200KB for a catalogue scan).
- **Fetched as an authenticated blob**, not `` — unlike `/api/images` this endpoint is scoped to the owner and an `img` tag can't send a bearer token.
- **Badge matches the existing corner-badge convention** used elsewhere in the app (see `ProductSourceBadge`) — rounded-full, theme-aware translucent background with a blur, standard corner spacing and z-index — rather than a one-off style. One component, reused everywhere the badge appears instead of copy-pasted.

## Testing

Backend: `test_collection_photos.py` and `test_collection_photo_normalize.py` for the photo endpoints and EXIF/XMP/ICC stripping, `test_own_photo_priority.py` for `has_scan_photo` propagation across every endpoint listed above (269 lines, covers the actual reported bug — the binder "add from collection" picker showing card backs for cards that had photos). Full backend suite (523 tests) passes.

Frontend: `collectionPhoto.test.js` regression-guards a real upload-path bug found along the way (client was sending the upload as JSON instead of multipart, and a validation-error response was crashing the whole page instead of showing a toast).


lamiskin:
@Git-Romer I've rebased this PR on main after the bigger changes over the last few days, hopefully it fits better with the new card system.

Git-Romer:
Thank you for the work on this. Before I change anything, I want to make sure I understand your intended behavior correctly.

From the original “Why” section and earlier implementation, I interpreted the feature as an alternative to entering an external image URL when TCGdex has no artwork. However, the latest version makes the owner’s photo replace TCGdex artwork even when catalogue artwork exists.

Was your intention:

1. To use the uploaded image only as a fallback when TCGdex has no image, or
2. To let users replace existing TCGdex artwork with a photo of their physical card?

My current interpretation and proposed upstream behavior is:

- TCGdex artwork has priority.
- A private uploaded image is used only when TCGdex has no artwork.
- The standard card back is used when neither exists.
- If TCGdex adds artwork later, the uploaded fallback remains stored privately but is no longer displayed.
- One uploaded fallback belongs to the user and card, so all quantities and variants use the same image.
- Scanner photos are retained only when the matched card has no TCGdex image and no existing fallback.
- Catalogue cards using a fallback show the camera badge.
- Manually created cards show only the pencil badge, never both badges.
- Each user gets a Settings action to permanently delete their uploaded fallback images and associated records.

If your intention was option 2, I would not merge that behavior into the main application as it currently works. I want catalogue cards to remain visually consistent throughout PokéCollector. I would instead consider showing the user’s physical-card photo separately inside the card dialog, possibly in its own tab, while keeping TCGdex artwork as the primary image everywhere else.

You are, of course, welcome to keep the alternative replacement behavior in your fork, pull the upstream changes I make here, and maintain its compatibility with future PokéCollector updates. I simply do not think replacing catalogue artwork is the right default behavior for the general application.

Could you clarify whether you intended the upload as fallback artwork or as a separate photo of the physical copy?

lamiskin:
Good question, and thanks for thinking it through carefully.

To answer directly: my original intent was closer to option 1 — cover the cards with no TCGdex image. While building it, I decided it'd be nice to also show our own photo even when TCGdex has a scan, since a photo of the actual physical card (condition, centering, etc.) has some value beyond the reference scan. That's how it ended up always-overriding rather than fallback-only — not an accident, but a scope creep from the original goal.

I like your suggestion of keeping the photo separate (its own tab in the card dialog) rather than replacing the primary image everywhere — that keeps catalogue cards visually consistent, which I agree matters more for the general app than it does for me personally.

Rather than picking one behavior as the fixed default, though — what if it were a setting? Default to your proposed behavior (TCGdex artwork primary, photo in its own tab), but let a user opt into "my photo overrides the catalogue image everywhere" if they want it. That way the app stays consistent out of the box, but people who want their own photos front-and-center (like me) have the option.

You are right though, if it doesn't fit the project, I am happy to maintain a personal fork.

Happy to rework the PR along those lines if that sounds reasonable — separate tab by default, with a per-user toggle for the override behavior.

Git-Romer:
Thank you, that clarification helps. I agree that a per-user preference can support both use cases.

I’ll take over the implementation from here and adapt the PR to the following behavior:

• By default, TCGdex artwork remains the primary image. If a personal photo exists, it is available in the card popup. When TCGdex has no artwork, the personal photo becomes the fallback.
• With “Prefer my card photos” enabled, the personal photo becomes the primary image everywhere that collection item appears. The TCGdex artwork remains available in the popup as the catalogue reference.
• If the preferred image is unavailable, the other image is used automatically.
• If neither exists, the standard card back is shown.
• Changing the setting only changes presentation; it never deletes or overwrites either image.

I’ll also handle the current conflicts with main and review the scanner-photo retention and cleanup behavior as part of the implementation.

Thanks again for the original work and for explaining how the always-override behavior developed. I’ll let you know when the revised version is ready for review.

PR #323, opened 2026-08-07
feat: batch scanner recognition with a persistent queue and Gemini rate limiting

lamiskin:
**Depends on #317.** This branch is built directly on top of it, so the diff below currently includes that PR's single commit too — it'll shrink to just this branch's own 13 commits automatically once #317 merges (real merge commit history, no rebase needed on my end).

## Why

The scanner recognized one photo per request, synchronously, with no rate limiting: a large upload held an HTTP request open past proxy timeouts, blew through Gemini's ~6/min free-tier limit, and a 429 failed the item outright with no retry.

## What

- **Extended recognition prompt**: `number_local`/`number_total` as separate fields (replacing an anchored-regex-parsed combined field that silently dropped the denominator), plus a `printed_total` vs. `Set.printed_total` mismatch flag and a `set_code` tie-break in ranking — both accuracy-tested against real cards before implementing.
- **`POST /cards/recognize/batch`**: uploads grouped into composite 2×2 grids per Gemini call (~4x fewer image tokens, locked in after grid-size accuracy testing), with a `singles` override to bypass batching for cards where it measurably hurts (vintage/non-Latin script).
- **Persistent queue** (`scan_jobs`/`scan_job_items`): a batch survives a restart and gets reviewed whenever, instead of living only in the HTTP response.
- **Shared Gemini rate limiter**: one process-wide token bucket so queued batches and interactive scans share a quota; a 429 now retries instead of failing outright (with a test-caught subtlety — the backoff penalty must not let tokens refill mid-backoff, or the bucket resumes full and immediately re-triggers the 429).
- **Daily vs. per-minute quota handling**: a daily-quota 429 now backs off an hour instead of retrying every 30s for six minutes straight — verified against a genuinely exhausted daily quota.
- **Ranking also uses** `printed_total`, `regulation_mark`, artist/HP, and a pHash re-rank gated behind a confidence margin (benchmarked: phash put the correct card first in 8/9 real photos across pools up to 62 candidates; only fires when unambiguous).
- **Dedicated `/scans` queue page** with an outstanding-item nav badge; the scanner modal is now capture-only for batches (single-photo scans still answer inline).
- **Structured scan traces** (opt-in via `SCAN_TRACE_DIR`) plus an offline accuracy analysis script, so future accuracy work has real data instead of guesses.

## Testing

Backend: 370 tests passing, verified in isolation at this branch's tip (separate from the accuracy/UI/lookup work stacked on top of it). Frontend: vite build clean and 101 vitest tests passing per the original development session, verified against a real containerized stack — I don't have node/npm in this environment to re-run that myself, so that reflects the original verification, not something I re-checked just now.

Two things are explicitly unproven on live data because the free Gemini daily quota ran out mid-development (noted inline in the relevant commits): the daily-quota recovery happy path, and scan-trace population on a real successful scan (traces are unit-tested and confirmed on a simulated run, not a populated live case).


Git-Romer:
Hi @lamiskin, quick update: I’m actively working through PR #323 in smaller, testable checkpoints. Thanks again for the original implementation.

I’ve pushed the first checkpoint, covering:

• Secure persistent background queue and restart recovery
• Image validation, upload limits and 14-day retention
• Fair scheduling, retries and concurrency protection
• Batch staging from camera or gallery
• Progress, navigation and review workflows
• Photo/candidate comparison, retry, dismiss and add flows
• The captured-photo preview for single scans

Composite scanning, expanded prompts/ranking, pHash and diagnostic tracing remain inactive for now. I’ll first validate the current queue and interface.

Afterwards, I’ll work through the optional 2×2 composite flow and individual fallback, followed by the remaining matching improvements. The PR is still actively being worked on and isn’t ready for final review yet. Thanks for your patience!

Git-Romer:
## Final implementation update

The original description no longer reflects the full scope of this branch. I expanded and hardened the scanner with the following changes:

- Unified individual and batch scanning into one staging and persistent review flow. Every submission, including a single photo, now enters the same restart-safe scan queue.
- Added device camera and gallery staging for up to 50 photos, per-photo individual-processing overrides, exit/resume support, and an optional positioning guide.
- Added two-to-four-card composite recognition to reduce Gemini requests, with automatic individual fallback only for missing or uncertain positions.
- Added fair background dispatch, leases, retry limits, 14-day expiry, private sanitized image storage, and cleanup after confirmation, dismissal, job deletion, or expiry.
- Added database-backed scan and quota state with upgrade migrations for existing installations.
- Added shared per-key Gemini quota coordination. Structured daily quotas are distinguished from short-term throttling, provider retry delays are honored exactly, and quota retries do not consume recognition attempts.
- Expanded Gemini extraction and deterministic matching with split collector number, printed total, set code, regulation mark, language, artist, HP, and explicit handling of uncertain small text. Candidate lookup uses the detected language with an English fallback.
- Added conservative local perceptual-hash matching for exceptionally clear visual matches. It never overrides confirmed metadata contradictions and safely falls back when ambiguous.
- Added persistent scan review with ranked candidates, retry reasons and countdowns, individual retry, dismissal, collection-add confirmation, and navigation badges.
- Added optional consent-controlled scanner diagnostics. Server support and per-user consent are both required; consent is off by default. Disabling consent stops future capture without deleting existing data, while a separate action deletes stored diagnostics. Trace files are user-scoped, private, credential-redacted, and removed with account deletion.
- Added an offline diagnostics analyzer for retrieval, ranking, metadata, perceptual-hash, and final-decision accuracy.
- Updated the README, environment example, and architecture, backend, and frontend documentation to reflect the final scanner behavior and privacy controls.
- Bumped the release version to `1.34.0`.

All required checks pass.

PR #324, opened 2026-08-07
fix: scanner accuracy — orientation retry, Energy card symbols, artwork-hash fallback

lamiskin:
**Depends on #323** (and transitively #317). Built directly on top of it, so the diff currently includes both PRs' commits — will shrink to just this branch's own 4 commits once they merge in order.

## What

Four fixes found by running the scanner against a real 72-card set, not by reading the code:

- **Retry other orientations when a photo yields no card name.** A card photographed upside down returned an empty name — the only thing catalogue search has to go on — so the scan failed outright even though every other field was readable. Now retries at 180/90/270° when a read produces no name, bounded so an unreadable photo can't spin.
- **Identify basic Energy cards by their symbol, not their printed name.** A basic Energy prints "Basic Energy" or just "ENERGY"; TCGdex names the same card "Water Energy". The printed name matched nothing, so the scan fell back to whatever ranking happened to return — a Water Energy could come back as a Darkness Energy. Now asks the model for the elemental type (only ever shown as a symbol, never text) and searches `"<Type> Energy"` when the printed name is generic. Also fixed a general candidate-cap bug this uncovered: results were floated by recognized number but not by recognized set code, so a photo with a readable code and no readable number kept whatever order TCGdex happened to return. Measured: energy exact-match went 4/6 → 6/6, set self-consistency 67/72 → 70/72.
- **Match artwork by hash ensemble before falling back to Gemini.** The Gemini visual-verification fallback was measured dead weight (0/290 recorded scans decided by it, and it silently 403'd without an API key). Replaced with a weighted phash+dhash+colorhash ensemble, benchmarked at 96.9% on 64 cards with confirmed answers vs. 98.4% for SIFT (which needs OpenCV) — equal on what matters for zero new dependencies. Thresholds are calibrated from the actual score distribution (not guessed) and pinned by tests.
- **Store scanned photos the right way up.** Recognition tolerates rotation, so nothing upstream noticed a sideways photo — the review UI showed it sideways beside upright candidates. The model can't self-report orientation (0/18 correct on rotated test cards), but once a card is matched, its upright catalogue scan can: rotating the photo four ways and comparing recovered the angle exactly (36/36 across 9 cards). Only applied when confident.

## Testing

Backend: 406 tests passing, verified in isolation at this branch's tip.


lamiskin:

PR #325, opened 2026-08-07
feat: overhaul the scan review comparison — instant candidate loading, linked zoom, batch accept, straightening

lamiskin:
**Depends on #324** (and transitively #323, #317). Built directly on top of it, so the diff currently includes all three PRs' commits — will shrink to just this branch's own 15 commits once they merge in order.

## What

A series of fixes and features to the side-by-side photo/candidate comparison view, in the order they were found:

- **Instant candidate loading**: expanding a candidate used to take ~5s (a cold fetch from the TCGdex asset CDN, timing out inside recognition too). Recognition now pre-caches the top two candidates' full-res scans; a new endpoint caches on miss so it's slow exactly once, ever, per card (measured: 5,032ms cold → 8.5ms cached). A blurred low-res thumbnail cross-fades to the real image while it loads instead of showing a blank frame.
- **Batch accept from the comparison view**: reviewing used to mean expand → close → click tile → fill form → find the next photo. The comparison now carries the candidate's details with an Accept button, walks straight to the next photo on accept, and arrow keys step through a photo's candidates with prefetching so stepping doesn't flash empty.
- **Linked zoom/pan** shared between both images (so zooming the photo brings the candidate with it — comparing is the point), with scroll-toward-pointer, drag-to-pan at the correct 1:1 tracking speed (an earlier version panned at ~1/6 speed from two compounding transform-math bugs, since fixed), and click-to-zoom-at-point as a discrete alternative to the wheel.
- **Manual + automatic straightening** for the ~7% of matches with no catalogue scan to compare rotation against (energy cards are over-represented here). Landscape vs. portrait alone gets a photo onto the right axis (98% on 80 synthetic rotations); a manual quarter-turn control covers what automatic detection can't safely resolve — two measured automatic heuristics (61%, 58%) weren't reliable enough to trust unsupervised, since a wrong auto-rotation is worse than none.
- **Collapse confirmed items** to a single line instead of leaving a large empty panel once the photo is dropped post-resolve, keeping the review list a list of what's still outstanding.
- A run of smaller alignment/sizing/spinner fixes along the way (candidate and photo rendered at mismatched scale, misaligned baselines, a spinner in the wrong place, a stale pre-rotation photo left on screen after the backend straightened it server-side).

## Testing

Backend: 414 tests passing at this branch's tip. Frontend: 101 vitest tests passing per the original development session (verified against a real containerized stack) — not independently re-run by me in this session, no node/npm available here.


lamiskin:

PR #326, opened 2026-08-07
feat: external lookups (image/eBay/Cardmarket/Pokéllector search, Lens) when there's no catalogue scan

lamiskin:
**Depends on #325** (and transitively #324, #323, #317). Built directly on top of it, so the diff currently includes all prior PRs' commits — will shrink to just this branch's own 7 commits once they merge in order.

## Why

Roughly one match in 14 has no image in TCGdex — trainer kits, Japanese printings, the newest energy subsets — which is exactly where a reviewer most wants to look something up, and exactly where the comparison view (#325) had nothing to show.

## What

For candidates with no catalogue scan, offers:

- **Image search, eBay, Cardmarket, and Pokéllector** links built from the card's name/set/number — links only, nothing fetched or stored, so no licensing question and no third-party dependency. (Two alternatives were checked and ruled out for coverage gaps: pokemontcg.io stops trainer kits at 2006 and is English-only; Bulbagarden's titles can't be reliably derived from a TCGdex id.) Search strings carry the set *name* as well as its code, since that's what a listing caption actually says.
- **Reverse image search on the user's own photo**, via Google Lens — the most direct way to identify a card no catalogue has at all. This went through three failed attempts before landing: clipboard-based paste corrupted the image at every size/format tried (traced to the canvas `toBlob` step, not the encoding itself — verified the canvas pixels were correct via `getImageData`), so the final version hands Lens the original stored JPEG as a direct download with no re-encoding step, since a silently-wrong image is worse than a slightly-larger one for a tool whose entire job is confirming identity.
- **Language-aware Pokéllector search** — its Japanese subdomain still indexes by English name, so it uses the model's English reading rather than the printed one.
- **Candidate language badge fix**: the existing badge only distinguished German-or-English, so every Japanese/French/Korean printing displayed as English — worst exactly where language is often the only thing distinguishing two candidates. Now shows the actual language from the full TCGdex language list.

## Testing

No new backend surface — this is frontend-only, so the backend suite (414 tests) is unchanged from #325. Frontend behavior as verified in the original development session against a real containerized stack — not independently re-run by me in this session (no node/npm here).


lamiskin:

PR #336, opened 2026-08-09
Prioritise alphanumeric and suffixed collector numbers

### LINKED ISSUE #335: Scanner cannot prioritise trainer gallery or promo card numbers (TG01, SV107)
Since #317 the scanner prioritises candidates whose collector number matches the recognised one, which fixed correct matches being truncated away by the candidate cap.

That prioritising only sees numbers that *start with digits*, so cards whose printed number is alphanumeric are still lost, and suffixed numbers can match the wrong card.

### Steps to reproduce

Recognise a card whose printed number has no leading digits - a trainer gallery card such as `TG01`, or a promo such as `SV107`:

```python
>>> from api.recognize import normalize_scanner_card_number
>>> normalize_scanner_card_number("TG01")
None
>>> normalize_scanner_card_number("SV107")
None
>>> normalize_scanner_card_number("74a")
'74'
```

### Expected

- `TG01` prioritises the card numbered `TG01`.
- `74a` prioritises the card numbered `74a`, and not the card numbered `74`.

### Actual

- For `TG01` and `SV107`, `prioritize_cards_by_number` returns the list unchanged and reports 0 matches, so the right card stays wherever the API happened to place it. If that is below the 8 candidate cap it is dropped - the same truncation #317 set out to fix, still reachable for these numbers.
- `74a` normalises to `74`, which is a different, real card, so a suffixed number can prioritise the wrong print.

### Notes

`test_recognize.py` asserts `normalize_scanner_card_number("TG01") is None`, so this reads as a known limit of that helper rather than an oversight in #317 - hence an issue rather than a bug report against it.

Happy to open a PR; I have one prepared with tests.

jasonbayton:
## What and why

Fixes #335.

Scanning a trainer gallery or promo card (`TG01`, `SV107`) could not find it. `prioritize_cards_by_number` matches on leading digits, and those numbers have none, so nothing was prioritised and the card stayed below the candidate cap - the truncation #317 fixed, still reachable for these numbers.

In the other direction the rule was too loose: `74a` reduced to `74`, which is a different, real card.

Both sides of the comparison now go through `card_number_variants`, so `63/100` and `063` still meet on `63`, `TG01` matches `TG01`, and `74a` no longer matches `74`. `normalize_scanner_card_number` keeps its existing contract and callers, so its behaviour and its tests are unchanged.

`printed_number_variants` and `card_number_variants` are added to `services/card_numbers.py`, alongside the existing number helpers. They drop the set total and offer both zero-padded and unpadded forms for purely numeric values, because TCGdex is inconsistent between sets - Base Set Charizard is localId `4` while Phantasmal Flames Charmeleon is `012`.

## Validation

- [x] Relevant automated checks pass - full backend suite on this branch: 407 tests, no new failures. (`test_accent_search` reports the same 5 errors on an untouched `v1.33.2` checkout, so they are unrelated to this change.)
- [x] Documentation was updated when behavior changed - no user-facing docs describe scanner number matching; the behaviour is documented in docstrings and tests.

## Card UI (when applicable)

Not applicable - backend only, no UI change, so screenshots are not relevant.

## Tests

- `test_scanner_number_matching.py` - `TG01` and `SV107` prioritised; `74a` does not match `74` and vice versa; the existing leading-zero and set-total behaviour preserved; unreadable numbers still leave the order untouched.
- `test_card_numbers.py` - variant generation, including that a suffix is never reduced away.

Each new assertion was checked against the pre-fix implementation: restoring the digits-only match fails the five new behavioural tests while the existing 23 scanner tests stay green.

### DIFF
--- a/backend/api/recognize.py
+++ b/backend/api/recognize.py
@@ -11,6 +11,7 @@ from sqlalchemy.orm import Session
 from api.auth import get_current_user
 from database import get_db
 from models import Setting, UserSetting, User, Set
+from services.card_numbers import card_number_variants
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


jasonbayton:
Closing this per your note in #342.

The matching improvements in #323 cover the collector-number handling, and the branch conflicts with them now, so carrying it further would just duplicate merged work.

Thanks for the detailed reply and the roadmap.

PR #337, opened 2026-08-09
Local-first recognition, collection provenance, and design polish

groovecityJO:
## What changed

Bundles the remaining John John's PC refinement work sitting on `fix/design-system-tokens`.

- **Local-first recognition** — recognition and the agent run against local data first; adds scan history plus `test_recognize_local_first` / `test_agent_local_only` coverage. `backend/api/recognize.py` is substantially trimmed as part of the rework.
- **Collection provenance & care** — cards track provenance, care, and creator/lot media; sealed products and binders stay collection-first (`Collection.jsx`, `InventoryIntakeModal.jsx`, models/schemas).
- **Surfacing** — Analytics visualize view, plus chase-card and first-release surfacing across Collection, Sets, SetDetail, Products, Discover, and Settings.
- **Design system** — token pass on `archive.css` and page layouts (the branch's original focus).
- **i18n** — full-locale copy for chase-card and first-release strings, with `chaseCardsLanguage` / `firstReleaseLanguage` tests.
- **Windows launcher** — `scripts/windows/` install + open scripts and the app icon/PNG assets.
- **Docs** — implementation plan and execution checklist under `docs/superpowers/plans/`.

## Testing

New frontend and backend test files accompany each feature area (see the `*.test.jsx` / `test_*.py` additions). Reviewers should run the frontend and backend suites.

## Reviewer notes

- Opened from a fork (`groovecityJO/pokecollector`) because the author account has read-only access to this repo.
- Session scratch artifacts (`.audit/`, `.review-head-*.txt`, `design-qa*`) were deliberately left out of the commit.
- 121 files, +4006 / -1431.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #338, opened 2026-08-09
Add Windows launcher for the local collection stack

groovecityJO:
## What

Adds a one-click Windows launcher for running John John's PC locally. Salvaged as a focused, self-contained change onto current `main` (part of breaking up the stale #337).

- **`scripts/windows/Open-John-Johns-PC.ps1`** — waits for Docker Desktop, brings the stack up via `docker-compose.yml` + `docker-compose.local.yml`, then opens the archive at `http://127.0.0.1:13000` once it responds. Shows friendly dialogs on failure; never changes Windows services automatically.
- **`scripts/windows/Install-John-Johns-PC-Launcher.ps1`** — creates Desktop and Start Menu shortcuts pointing at the launcher, with the icon.
- **`docker-compose.local.yml`** — a local-deployment override of the base compose (localhost-only ports, external postgres volume, local backup mounts). Overrides only existing `postgres`/`backend`/`frontend` services.
- **`frontend/public/john-johns-pc-launcher.{ico,png}`** — launcher icon assets.

## Why

`main` had no local one-click launch flow. This is additive and touches no application code.

## Testing / notes

- The override targets the same service names as the base `docker-compose.yml` and maps the frontend to `127.0.0.1:13000`, matching the URL the launcher polls.
- `docker-compose.local.yml` contains machine-specific volume paths appropriate for a personal local deployment; adjust for other machines.
- Opened from a fork (`groovecityJO/pokecollector`) as the author account has read-only access to this repo.


Git-Romer:
Thank you for separating the Windows launcher from #337 and for the work you put into making local startup easier.

After reviewing the complete implementation against current `main`, I do not think this PR can be merged as it stands. It is tied to one personal installation through the `John John's PC` branding, fixed Windows paths, drive assumptions, ports, and external database-volume name.

I also found several functional risks:

- A clean-install dry run fails because `pokecollector_postgres_data` must already exist.
- The override starts a separately named Compose stack while pointing PostgreSQL at the existing PokéCollector volume. This could conflict with an already-running standard installation.
- The backend volume override replaces the complete volume list. On current `main`, this removes persistent storage for Pokédex images, queued scanner uploads, and scanner diagnostics.
- Startup errors are hidden behind a generic dialog, and readiness checks verify only the frontend rather than backend health.
- There is no generic configuration or installation documentation, and the launcher has not been validated through a Windows-specific test or check.

I also think PokéCollector’s standard Docker setup is already reasonably accessible: getting started requires only a few commands, while the documentation covers optional configuration and hardening. Maintaining another installation path would add complexity and another deployment flow that must remain compatible with every future update.

Because resolving this would require redesigning and continuously maintaining the launcher—not merely making a small correction—I intend to close this PR. You are, of course, welcome to retain the launcher in your fork and adapt it to your personal environment.

If broader demand for a generic Windows launcher appears later, we can discuss a separate proposal designed specifically for PokéCollector and current `main`.

Thank you again for the contribution and for extracting this into a focused PR.

groovecityJO:
Understood, and agreed — closing is the right call. The launcher and `docker-compose.local.yml` are specific to a personal install (paths, branding, the external `pokecollector_postgres_data` volume), and you're right about the concrete risks.

In particular, the `volumes: !override` on the backend service is a real regression: it replaces the full volume list, so on current `main` it would drop the Pokédex image and scanner-upload volumes. That alone makes it unsafe as a general path, independent of the branding.

I'll keep it in the fork and adapt it there. If a generic Windows launcher is ever worth doing, it'd need to be designed against current `main` with additive compose overrides, backend health checks, and real docs — a separate proposal, as you say. Thanks for the careful review.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #339, opened 2026-08-09
Record recognition scan history

groovecityJO:
## What

Adds short-lived scan history for card recognition. Second focused salvage PR onto current `main` (breaking up the stale #337).

- **`models.py`** — new `ScanHistory` table (recognized name/number/language, match count, top match, `created_at`/`expires_at`). Auto-created via `create_all`; no migration needed.
- **`api/recognize.py`** — `purge_expired_scan_history` / `record_scan_history` / `serialize_scan_history` helpers; each successful `POST /cards/recognize` now records an entry and returns it as `scan_history` in the response; new `GET /cards/scan-history` returns recent non-expired scans (newest first, expired rows purged on read). Retention: 14 days.
- **`api/client.js`** — `getScanHistory` helper.
- **`CardScanner.jsx`** — shows a scan-expiry note from the recognize response.
- **`tests/test_scan_history.py`** — unit tests for recording and expiry purge.

## Why

`main` had no record of past scans. This is additive: a new table plus one new read endpoint, wired into the existing recognizer.

## Adaptation note

On the source branch this feature was dormant scaffolding — `record_scan_history` was only ever called from tests because recognition was disabled there. Here it's wired into main's live `recognize_card`, so it actually records scans.

## Testing

`tests/test_scan_history.py` passes (2/2) under SQLite:
```
tests/test_scan_history.py::ScanHistoryTests::test_purge_expired_scan_history_only_removes_expired_rows PASSED
tests/test_scan_history.py::ScanHistoryTests::test_record_scan_history_expires_after_retention_window PASSED
```
Ran in a local venv (Docker was unavailable in the authoring environment); the backend's normal test path is `pytest` in the backend image.

Opened from a fork (`groovecityJO/pokecollector`) as the author account has read-only access.

### DIFF
--- a/backend/api/recognize.py
+++ b/backend/api/recognize.py
@@ -1,18 +1,78 @@
 import base64
 import asyncio
+import datetime
 import httpx
 import os
 import json
 import re
 from services.tcgdex_languages import is_supported_tcgdex_language, normalize_tcgdex_language
 import logging
-from fastapi import APIRouter, UploadFile, File, HTTPException, Depends
+from fastapi import APIRouter, UploadFile, File, HTTPException, Depends, Query
 from sqlalchemy.orm import Session
 from api.auth import get_current_user
 from database import get_db
-from models import Setting, UserSetting, User, Set
+from models import Setting, UserSetting, User, Set, ScanHistory
 
 logger = logging.getLogger(__name__)
+
+SCAN_HISTORY_RETENTION_DAYS = 14
+
+
+def purge_expired_scan_history(db: Session, user_id: int | None = None, now: datetime.datetime | None = None) -> int:
+    """Delete scan-history rows past their retention window; returns rows removed."""
+    now = now or datetime.datetime.utcnow()
+    query = db.query(ScanHistory).filter(ScanHistory.expires_at < now)
+    if user_id is not None:
+        query = query.filter(ScanHistory.user_id == user_id)
+    return query.delete(synchronize_session=False)
+
+
+def record_scan_history(
+    db: Session,
+    *,
+    user_id: int,
+    source_reference: str | None,
+    recognized: dict,
+    matches: list[dict],
+    now: datetime.datetime | None = None,
+) -> ScanHistory:
+    """Persist a recognition attempt as a short-lived scan-history entry."""
+    now = now or datetime.datetime.utcnow()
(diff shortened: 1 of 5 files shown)


Git-Romer:
Thank you for working on this. The idea of retaining compact scan results for a limited time is useful, but the scanner architecture has changed substantially since this PR was opened.

PR #323 introduced persistent scan jobs and items that already store recognition results, candidates, ownership, timestamps, and expiry information for both direct and queued scans. Merging this PR now would create a second scan-history system that only covers successful direct scans.

I also noticed that the new history endpoint is not currently used by the frontend, so users would not have an actual history interface. Cleanup only occurs after another successful scan or a request to that unused endpoint, meaning records may remain beyond the intended 14 days. Account-deletion cleanup and broader scanner-flow tests are also missing.

Because of this, I do not think updating this implementation to current `main` is the right direction. If we add completed scan history later, I would build it on top of the existing `ScanJob` and `ScanJobItem` lifecycle so direct scans, queued scans, failures, retention, account cleanup, translations, and the user interface all use one consistent system.

Does that assessment make sense to you, and is there any part of your intended use case that the scanner introduced in #323 does not already support? Unless I am missing something important, I intend to close this PR as superseded rather than ask you to resolve the current conflicts.

groovecityJO:
That assessment makes sense, and I agree — I don't see a use case here that `ScanJob`/`ScanJobItem` from #323 doesn't already cover. Those models already persist `recognized` + `matches` with per-job `expires_at`, ownership, and timestamps, across both direct and queued/batch scans, so this `ScanHistory` table is a narrower, redundant second system.

You're also right about the concrete gaps: the new endpoint isn't wired into the frontend, cleanup only runs on the next scan or a hit to that unused endpoint (so rows can outlive 14 days), and there's no account-deletion cleanup or broader flow tests.

Context on how this happened: this PR was salvaged from a stale branch where recognition was disabled and this history table was dormant scaffolding (only exercised in tests). Adapting it onto current `main` without building on `ScanJob` was the wrong call given #323.

Closing as superseded — agreed that completed scan history, if wanted, belongs on top of the `ScanJob`/`ScanJobItem` lifecycle so direct/queued scans, failures, retention, account cleanup, translations, and the UI all share one system. Thanks for the thorough review.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #346, opened 2026-08-10
fix: recover Energy cards and set-code-only matches dropped from search

lamiskin:
## What and why

To improve scanner recognition accuracy.

Two real accuracy gaps in the scanner's matching layer, found by replaying 65 ground-truth-labeled scan traces through the matcher offline (already-extracted card_info + the real card/set DB + live TCGdex — no Gemini/Ollama calls needed to test this layer):

- **Generic Energy cards search for nothing.** A card printed as "Basic Energy" or "ENERGY" doesn't match TCGdex's actual catalogue names ("Water Energy" etc). The elemental symbol now substitutes in the searched name when the printed name is generic — same idea `select_search_candidates` already applies to a misread number.
- **A correct set code can't rescue a card from search truncation.** TCGdex sorts name-search results by card number, so anything past the baseline slice is silently dropped once a set has enough prints — even when the set code was read correctly. `select_search_candidates` only ever floated matches by number; it now also floats by recognized set code.

Result on the 65-trace benchmark: top-1 accuracy 62/65 → 65/65.

## Validation

- [x] Relevant automated checks pass — full backend suite: 502 passed, 8 pre-existing skips, 0 failures; end-to-end 65-trace benchmark 65/65
- [x] Documentation was updated when behavior changed — `docs/ARCHITECTURE.md` Scanner Flow section

New unit tests added: set-code float in `select_search_candidates` (`test_recognize.py`), energy-name substitution (`test_card_field_cleanup.py`).

## Card UI (when applicable)

N/A — backend-only change, no card UI involved.


Git-Romer:
Thank you for working on this. Before we continue with the implementation, I would like to clarify the intended outcome of the Energy-card change.

As I understand it, recognizing the Energy symbol changes a broad search such as Basic Energy into a narrower search such as Grass Energy. However, there are many different Grass Energy printings, so the symbol alone cannot identify the exact card.

Do you expect this change to:

1. identify the exact Energy-card printing automatically and confidently, or
2. only narrow the candidate list by Energy type so that another step—set code, collector number, artwork comparison, or manual selection—can determine the exact printing?

If exact automatic matching is expected, what identifies the specific printing when the set code and collector number are missing or unreadable?

Could you also attach photos of some of the actual Energy cards that failed during your original scan? For each photo, please provide:

• expected language;
• expected set name or set code;
• expected collector number;
• what PokeCollector selected before this change;
• what PokeCollector selects with this PR.

Ideally, please include at least one difficult example where the set code or collector number is unreadable. My available cards show those identifiers clearly, which makes them relatively easy for Gemini and does not properly demonstrate the benefit of recognizing the Energy symbol.

I tested a German SVE 001 Grass Energy. Gemini correctly recognized SVE, 001, and the Grass type, but PokeCollector still matched it to PAL 278. Because set code and collector number are stronger identifiers than the Energy symbol, the PR should not lose or override them.

A complete end-to-end example would help establish the intended behavior:

1. current code selects the wrong card;
2. Energy-type recognition narrows the search appropriately;
3. the correct candidate enters the candidate list;
4. another reliable signal selects the exact printing—or the UI presents the narrowed candidates for manual selection.

Once the expected behavior and real examples are available, I can retest the PR against the same cards.

lamiskin:
Thanks for the thorough review, and especially for catching the SVE 001 → PAL 278 regression — that's a real problem with letting the energy symbol take priority over a readable set code/number, and it needs to be fixed properly rather than patched around.

I'm going to close this for now rather than keep it open while I gather the kind of end-to-end evidence you're asking for. If the underlying matching gaps show up again, I'll come back with better validation data (including unreadable-identifier cases) and open a fresh PR rather than resurrect this one.

Appreciate the detailed look.

PR #347, opened 2026-08-10
fix: recover pHash's declined matches with a colour-aware artwork pass

lamiskin:
## What and why

To improve scanner recognition accuracy.

The primary pHash stage is deliberately strict and abstains rather than guess when candidates are too close or too far. Replayed against the same 65 ground-truth traces used in #346, it declined on 9 of 58 scoreable cases. This adds a second pass on exactly those declines: phash + dhash + colorhash, colour double-weighted — colour is what separates same-artwork reprints that a single structural hash cannot.

Reimplemented phash/dhash/colorhash from scratch in numpy/PIL instead of importing the `imagehash` library, the same way recognize.py's own primary pHash stage already avoids that dependency. Verified bit-for-bit identical to imagehash's actual output across 15 real card photos (cropped and uncropped, 90/90 checks) before wiring it in.

Independent of #346 — applies cleanly on top of `main` alone.

Result: of the 9 traces where the primary pHash stage declined, this resolves 3 — all correct, zero wrong answers. Left as an honest number in the module docstring rather than a stronger claim from a narrower internal benchmark.

## Validation

- [x] Relevant automated checks pass — full backend suite: 502 passed, 8 pre-existing skips, 0 failures; all three hand-rolled hashes verified bit-for-bit identical to `imagehash`'s real output (90/90 checks); end-to-end 65-trace benchmark re-run after wiring in, no regressions
- [x] Documentation was updated when behavior changed — `docs/ARCHITECTURE.md` Scanner Flow section

New tests added: `test_card_image_match.py` (ensemble distance, best-match selection, rotation detection, calibration).

## Card UI (when applicable)

N/A — backend-only change, no card UI involved.

### DIFF
--- a/backend/api/recognize.py
+++ b/backend/api/recognize.py
@@ -12,6 +12,7 @@ from email.utils import parsedate_to_datetime
 from functools import lru_cache
 from urllib.parse import urlparse
 from services.tcgdex_languages import is_supported_tcgdex_language, normalize_tcgdex_language
+from services import card_image_match
 from services.gemini_rate_limit import (
     GeminiKeyBlockedError,
     acquire_gemini_slot,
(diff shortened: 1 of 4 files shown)


Git-Romer:
Thanks for working on this! I would like to better understand how well it performs during normal PokeCollector usage.

Could you share the original, unprocessed card photos you used for validation, along with their expected card IDs and candidate pools? I am particularly interested in the nine scans rejected by the existing pHash step and the three recovered by this fallback.

I tested the PR with six real handheld photos previously submitted to PokeCollector. The existing pHash identified one, but the new fallback recovered none of the remaining five. Rotation, perspective, distance, and framing appeared to affect the results considerably.

Since the implementation uses a fixed artwork crop without detecting, aligning, or correcting the perspective of the card first, users seem to need an almost perfectly upright, centered, and tightly framed photo before it provides any improvement.

What use case do you have in mind? Is this intended for ordinary scanner photos, manually cropped images, or a more controlled capture process? Do you see it as a meaningful improvement for typical PokeCollector users? I would like to evaluate it under the same conditions users will encounter in practice.

lamiskin:
Thanks for testing this against real handheld photos — 1/6 recovered isn't a result I'd want to ship, and the framing/perspective sensitivity you found points at a real gap in the approach (no alignment/correction before the crop).

Closing this for now rather than iterating in place. If pHash-decline cases come up as a real problem again, I'll revisit with a more robust approach and real validation data under normal usage conditions, and open fresh.

Appreciate you digging into it.

PR #354, opened 2026-08-10
Warn before enabling multi-user mode, add lockout recovery (USER_MODE + reset script)

### LINKED ISSUE #348: Enabling Multi-User Mode locks the admin out with no warning, no confirmation, and no password step
Enabling **Multi-User Mode** in Settings takes effect immediately, with no warning and no confirmation, and drops the operator at a login screen for an account whose password they have almost certainly never seen. There is no password reset command in the repository, so the only way back in is hand-crafting a bcrypt hash into Postgres.

This was already reported by a user in #81 ("I have enabled multi users in the settings and I am directly forwarded to a login screen ... I can not login neither did I see any setting for creating accounts"). That issue was closed as a `question` and answered with a log-scraping recovery plus a README line in #82. I would like to reopen the underlying problem rather than the symptom: documenting a recovery procedure does not stop the next operator walking into it, and the trap is still fully present on `main` today. A one-click control that can permanently lock an admin out of their own installation is a UX defect that warrants a code fix, not a docs note.

I am happy to raise a PR for this if you are open to the approach below.

## Reproduction

Verified on a clean `docker compose up` at `main` (`2bc0eef`, v1.33.2), fresh volume, no `ADMIN_PASSWORD` set.

```
1. GET  /api/auth/mode                     -> {"multi_user":false}
2. GET  /api/auth/me      (no token)       -> 200 {"username":"admin","role":"admin"}
3. GET  /api/collection   (no token)       -> 200

   --- flip the toggle, exactly what Settings.jsx sends ---
4. PUT  /api/auth/mode {"enabled":true}    -> 200   (no token sent)

5. GET  /api/auth/me      (no token)       -> 401 Not authenticated
6. GET  /api/collection   (no token)       -> 401
7. POST /api/auth/login   admin/admin      -> 401 Incorrect username or password
```

In the UI the toggle produces no dialog at all: `frontend/src/pages/Settings.jsx` calls `setAuthMode(val)` and then `window.location.reload()`, and `App.jsx` renders `<Login/>` because `!user && multiUser`.

## Why lockout is the default outcome rather than an edge case

`backend/services/auth.py::bootstrap_admin` seeds the admin with `secrets.token_urlsafe(20)` unless `ADMIN_PASSWORD` is provided, and logs it once at first boot:

```python
password = os.getenv("ADMIN_PASSWORD") or secrets.token_urlsafe(20)
```

A single-user operator is never shown a login screen, so they have no reason to record that password. By the time they discover the toggle, that log line has typically rotated away, and `ADMIN_BOOTSTRAP_LOG=false` suppresses it entirely. The README mentions only, under *Login*, that a random password "may be logged during bootstrap", which is a long way from the toggle that makes it load-bearing.

## Second problem: the mode can be flipped anonymously

Note step 4 above carried **no token**. In single-user mode `get_current_user` returns the admin for any untokened request, so the endpoint that enables authentication is itself reachable without authentication. On an internet-facing single-user installation, any anonymous visitor can flip the instance into multi-user mode and lock the owner out of it.

There is a second path to the same state, which any fix needs to cover as well:

```
PUT /api/settings/ {"multi_user_mode":"true"}   (no token)  -> 200
GET /api/auth/mode                                          -> {"multi_user":true}
```

`multi_user_mode` is in `ADMIN_ONLY_KEYS`, but in single-user mode "admin only" is satisfied by everyone, so the generic settings endpoint sets it just as readily as `/auth/mode`. A guard placed only on `/auth/mode` would be trivially bypassable.

I appreciate that single-user mode serves everything as admin by design, so this is not a privilege escalation in itself. The distinction is that this particular setting is the one that can permanently deny the real owner access, so it should not be settable by an unauthenticated request.

## Proposed behaviour

Make enabling multi-user mode require an admin password in the same operation, so lockout becomes structurally impossible rather than merely documented:

1. `PUT /api/auth/mode {"enabled": true}` without a password returns `400`, telling the caller a password is required.
2. `PUT /api/auth/mode {"enabled": true, "admin_password": "..."}` sets that password on the admin account, enables the mode, and returns an access token in the response.
3. The frontend shows a confirmation dialog before any of this, naming the admin username that will be required, and requires the password to be entered and confirmed.
4. The frontend stores the returned token, so the admin stays signed in instead of being bounced to a login screen. The reload-to-login disappears.
5. `multi_user_mode` is rejected on the generic `PUT /api/settings/` path, so `/auth/mode` is the single entry point that the guard covers.

This also closes the anonymous-flip hole as a side effect: once the request has to carry a credential the caller sets, a passing stranger can no longer trigger it.

Disabling multi-user mode is unaffected, and installations that set `ADMIN_PASSWORD` are unaffected beyond seeing the confirmation step.

Happy to adjust the shape if you would rather keep the flow closer to what exists today, for example a confirmation dialog only. I would just want to avoid landing something that still permits the lockout, given #81 shows it does happen to real users.

jasonbayton:
Fixes #348.

Enabling Multi-User Mode enforces the login screen immediately and signs the admin out, against an account whose password they usually never set (single-user mode never asks for one; the bootstrap password is random and only reaches the first-run log). There was no warning, no confirmation, and no reset tooling, so an operator could lock themselves out of their own installation. See #348 for the full report.

This keeps single-user mode exactly as it is (LAN, no login) and makes enabling it safe:

- **Warn before enabling.** The toggle now opens a confirmation dialog that states a login will be required immediately, names the account to sign back in as, and points to the reset script and the recovery variable. Enabling stays a plain flag flip, with no password step (an earlier draft set the password on this endpoint, which would have let an unauthenticated caller take the account over, so that approach was dropped).
- **`USER_MODE` recovery/override.** `USER_MODE=single` pins single-user mode and `USER_MODE=multi` pins multi-user, overriding the stored setting. `USER_MODE=single` is the way back in after a lockout: set it, restart, regain local access, reset the password, remove it. While it is set, `/api/auth/mode` reports `locked`, the Settings toggle is disabled with a note, and `PUT /api/auth/mode` returns 409. `USER_MODE=single` is logged loudly at boot as a local/LAN tool.
- **`scripts/set_admin_password.py`.** Set or reset an account's password from the host (`docker compose exec backend python -m scripts.set_admin_password`), prompted via getpass, the reset tooling the project did not have.

Docs cover setting a known password before enabling and recovering after a lockout.

## Scope

This fixes the lockout. It does not change that single-user mode is intentionally unauthenticated, and I would not want it read as doing so: a single-user install exposed to the internet is reachable by anyone regardless of this change. As noted on #348, the one thing worth doing alongside is making that expectation explicit in the docs (single-user has no auth, do not expose it; use multi-user behind the existing reverse-proxy guidance for public deployments) - that is a documentation call for you rather than something this PR assumes.

## Tests

`backend/tests/test_multi_user_mode_recovery.py` covers the plain flag flip, `USER_MODE` precedence in both directions, value parsing, the `locked` flag, the 409 while pinned, and that enabling never touches the admin password. The frontend translation-key check and build pass.


jasonbayton:
Thanks for finishing this off — the disable-side confirmation and the `multiUserSaving` reset were both gaps in what I sent, and the `warn_invalid` change is the right call for something called per request.

One thing the new guard misses, which I hit while merging this downstream.

`update_settings` (`PUT /api/settings/`) now refuses `multi_user_mode` with a 409, but `backend/api/settings.py` has a second writer — `set_setting` (`POST /api/settings/{key}`) — which takes any key, checks admin role, and writes it:

```
POST /api/settings/multi_user_mode
{"value": "true"}
```

That turns the login screen on or off with no confirmation and without going through `/api/auth/mode`, which is exactly what the 409 exists to prevent. `_apply_setting_side_effect` has no `multi_user_mode` branch either, so nothing compensates downstream.

`test_generic_settings_endpoint_cannot_change_user_mode` passes because it only exercises `update_settings`, so the suite reads as covering this when it does not.

It is admin-only, so this is not a privilege escalation — it is a way for an admin to disable authentication entirely without the warning the settings page is careful to show, which matters most on an install reachable beyond a trusted network.

I have a fix that shares one guard between both writers, keyed on a small table of settings that belong to a dedicated endpoint, so a third writer cannot quietly miss it, plus a test that fails against the current code. Happy to open it as a PR if useful.

PR #355, opened 2026-08-10
Add OpenAI-compatible vision providers to the card scanner

### LINKED ISSUE #342: Should the card scanner be provider-agnostic?
At the moment the scanner is built around Gemini specifically: the model helper, the URL builder, the error handling and the transient status codes all assume it. PR #323 adds rate limiting shaped the same way (`services/gemini_rate_limit.py`, keyed on a Gemini API key fingerprint).

I've been running a local fork with a small provider abstraction so the scanner can use OpenAI instead, because that's the key I already pay for. It's a `VisionProvider` base with two implementations, resolved from a setting, and it kept the existing Gemini behaviour intact. Roughly: the model, URL and error handling move into the provider, and `recognize.py` asks for whichever is configured.

Before I offer any of that upstream I'd rather ask whether you want it at all:

1. Do you want more than one vision provider supported, or is Gemini deliberate (free tier, known behaviour, one thing to support)?
2. If you do, is now the moment? The scanner work in #323 to #326 is substantial, and rate limiting in particular would want to be per-provider rather than Gemini-shaped. Deciding after that lands means reworking more.
3. If you don't, no problem, and it's useful to know so I can stop carrying it.

Happy to open a PR, or to leave it in my fork. I'd rather not cut across the work already in flight.

jasonbayton:
## Summary

Adds administrator-controlled OpenAI-compatible vision providers to the card scanner while keeping Gemini as the default.

This replaces the original free-text setup with a guided, guarded configuration flow. Administrators decide which providers, endpoints, and models are available. Users can then select only approved choices and provide their own key when required.

## What changed

- Added a provider-neutral scanner layer for Gemini and OpenAI-compatible vision APIs.
- Added administrator controls for provider availability, endpoint, labels, allowed models, and authentication requirements.
- Added an administrator-only custom-model test path.
- Made provider testing and saving atomic, so failed tests cannot partially change the active scanner configuration.
- Bound capability verification to the provider, model, and endpoint. Changing any of them requires a retest.
- Enforced capability proof for direct scans, new queued scans, and already queued work.
- Made disabled providers fail closed instead of silently routing scans to another provider; selecting a fallback requires an explicit test and save.
- Added genuine PNG capability fixtures, tolerant response parsing, and bounded transient retries.
- Added shared provider rate limiting and clearer authentication, billing, model, timeout, and availability errors.
- Prevented provider errors from being misclassified as model capability limitations.
- Fixed Gemini startup behavior so persisted per-user keys work without a server-wide startup key.

## Limited scanner mode

Some vision models can recognize one image but cannot compare the photographed card with multiple reference images.

When multi-image testing fails but single-image recognition succeeds, an administrator may explicitly acknowledge and save a limited configuration. In that mode:

- recognition, metadata matching, and perceptual image matching remain active;
- AI comparison against reference images is disabled;
- Settings and the card scanner display persistent warnings;
- changing the provider, model, endpoint, or key clears the approval and requires another test.

Limited-mode approval applies only to the tested user configuration. It does not silently enable a limited model for other users.

## Documentation and UI

- Added scanner-provider configuration documentation and updated the architecture, backend, frontend, README, and environment examples.
- Added translations for every supported language.
- Added desktop and mobile browser coverage for guided setup, custom models, test failures, limited-mode acknowledgment, warnings, and administrator-only details.

## Validation

- 695 backend tests passed, 8 skipped.
- 251 frontend unit and translation tests passed.
- 62 real Chromium browser tests passed across desktop and mobile.
- Production frontend build passed.
- Backend and frontend production Docker image builds passed.
- Version, Compose port, translation, and diff consistency checks passed.
- Fresh independent review completed on the current-main merged state.

Release version: `1.41.0`

Closes #342


Git-Romer:
Hi Jason, thanks again for starting this work. I’ve updated the PR branch with a substantially revised implementation based on current main.

The provider abstraction remains, but I reworked the configuration, error handling, security, documentation, and user experience:

- Gemini remains the default and keeps the same prompts, matching, automatic visual verification, queue lifecycle, warnings, and scan results.
- OpenAI-compatible providers are available only when explicitly enabled by the administrator.
- Users choose only approved providers and models from guided dropdowns - there are no user-editable endpoints or model names.
- Models are stored separately per provider, and the administrator’s default is marked as recommended.
- Configuration is tested and saved atomically using a real image-capability test.
- Rate limits support Retry-After, HTTP dates, OpenAI reset headers, persistent provider blocking, and bounded increasing fallback waits.
- Authentication, billing, missing-model, incompatible-response, and unsupported-request errors stop immediately instead of retrying for days.
- Provider responses and arbitrary API keys are protected from logs, API responses, queue errors, and scanner diagnostics.
- The Settings flow now includes readiness states, privacy/cost guidance, contextual setup links, friendly provider names, and collapsed administrator details.
- Setup documentation now covers Gemini, hosted OpenAI, Ollama, Docker/Compose networking, and generic OpenAI-compatible services.
- The complete scanner setup is localized across all supported language variants.

Could you please test the latest version, particularly Gemini and any hosted OpenAI or local compatible setup you have available? I’d also really appreciate your opinion on whether this revised guided flow feels right and whether anything valuable from your original approach should be restored or adjusted.

jasonbayton:
I'll have a look, on the face of it - 

- I don't see value in admin-gating the openAI compatible providers, that feels like a conflict to your earlier choice to allow users to choose their own models and provide their own keys. 
- Approved providers means the admin will be in and out updating that list, I explicitly avoided that, because again if my users are providing their keys, it's none of my business what they choose, and it shouldn't be my onus to ensure every potential model is available. I believe it also made it more difficult to allow for self-hosted LLMs. My opinion would change if admins were preventing users from using their own model/key and forcing a fallback to an admin/env key, then it'd make sense to gatekeep models, as the admin may be cost averse.
- Stopping immediately makes transient errors hard blockers. I'd soften to a few retries before quitting.

jasonbayton:
OK a deeper look: 

1. Headerless rate-limit backoff never resets after a success

provider_rate_limit.py:81 derives the next fallback penalty from the previously stored block, but nothing in scan_providers.py records a successful request or clears that history. A provider that emits intermittent 429s without reset headers therefore climbs 30s, 120s, 10m, 30m, 1h, 6h even when scans succeed in between.

Reproduced against the real limiter code on SQLite: a 30s block expired, a request succeeded, and the next headerless penalty was still 120s rather than back to 30s.

Compatible and self-hosted endpoints are the most likely to omit reset headers, so they absorb this disproportionately. Recording a success and resetting the escalation would fix it.

2. allow_visual_verification=True is now unconditional

recognize.py:1112 passes it unconditionally, and I couldn't find any provider or capability gate inside match_card_info.

The reason the original keyed the default off openai_requires_key() wasn't to add a preference, it was capability: hosted APIs handle the multi-image comparison well, and a self-hosted model usually doesn't, where the failure mode is a confident wrong pick rather than an honest miss. Since OPENAI_BASE_URL pointed at a local server is the main reason to want an OpenAI-compatible provider at all, that path now gets worse results than no verification.

3. The new bootstrap docs describe something that doesn't happen on a first start

docs/scanner-providers.md:49 and README.md:288 say GEMINI_API_KEY bootstraps the administrator's key on a new installation. main.py:63 calls init_db() before the administrator is created at line 68, so the import at database.py:989 finds no administrator on the first start.

Verified by calling the actual startup functions: 0 keys imported after the first start, 1 after a second. A fresh install reports that a key is required even though it's in .env, until it restarts again.

The ordering predates your commits, so this is really "the new guide promises current behaviour it doesn't have". Either the docs need a caveat or init_db() needs to run after the administrator exists.

4. Four of the new tests don't constrain what they look like they constrain

Mutation-tested against the full backend suite, all 658 tests still passing in each case:

- test_scanner_settings.py:181 only asserts the Gemini secret is hidden. Hiding gemini_api_key while exposing the newly added openai_api_key still passes.
- test_provider_rate_limit.py:73 has no live-state bystander. Purging every limiter row instead of only stale ones still passes, which would drop active blocks and resume hitting a provider early.
- test_provider_rate_limit.py:26 separates credentials but not endpoints. Dropping the endpoint from the scope fingerprint still passes, so two different keyless local endpoints would share one block.
- test_scanner_settings.py:168 checks an image part exists but not that it decodes. Replacing the payload with invalid base64 still passes.

Every other mutation I applied to the new backend tests failed its intended test, so this is a gap rather than a pattern. The per-provider settings test correctly caught an over-broad write to the Gemini bystander.

Minor

- Losing log_upstream_detail costs operators the only practical diagnostic on compatible endpoints, where "unsupported image format" or "context length exceeded" is often the whole answer. I wouldn't restore my version, since your reason for not letting provider-controlled text into logs is sound and your tests enforce it. Bounded structured fields, or a debug line with the configured credential explicitly scrubbed, would cover it.
- One dead key: scanner_visual_verification is in the blocklist in settings.py:80 but read nowhere, which will puzzle someone later if the capability default doesn't come back.

Git-Romer:
Hi Jason, thanks for the detailed review. I’ve pushed another update addressing your findings:

- Headerless rate-limit backoff now resets after a successful request. The reset is race-safe, so an older success cannot clear a newer 429.
- GEMINI_API_KEY is now imported correctly on the first startup.
- The four tests you identified now properly cover secret exposure, active limiter rows, endpoint separation, and valid image payloads.
- Every provider and model must pass a real two-image capability test before it can be saved. This lets self-hosted providers retain the complete Gemini-equivalent visual-verification workflow without assuming capability.
- Transient network and provider failures still receive bounded retries. Only permanent authentication, billing, model, and invalid-request errors stop immediately.
- Administrators can test arbitrary custom models under Advanced. Normal users continue to receive guided, approved choices.
- Provider fingerprints, Gemini error handling, diagnostics, and documentation received an additional security and compatibility review.

I deliberately kept endpoints administrator-controlled to avoid arbitrary server requests and SSRF risks. I also kept the guided model selection for normal users because PokéCollector should remain approachable for people who are not familiar with model names or provider configuration. Self-hosted services such as Ollama remain supported through administrator configuration and the tested custom-model path.

Could you please test the latest branch again, especially the custom-model and self-hosted flow? I’d appreciate your opinion on whether this compromise preserves the flexibility you intended while keeping setup safe and understandable.

jasonbayton:
Alrighty, I ran an OpenAI-compatible stub locally against the new branch

1. The image payload test - replacing both PNG fixtures with valid base64 that isn't image data still passes the entire backend suite, exit 0. The other three now correctly fail their mutations. The assertion needs to look at the decoded bytes rather than at base64 validity.

2. Capability proof isn't bound to the endpoint it was proved against
The verified-model record survives a change to OPENAI_BASE_URL. Test a model against endpoint A, save, repoint the base URL at endpoint B, and the stored verification still counts, so allow_visual_verification=True runs against a model that was never tested there.

Since the endpoint is administrator-controlled and set by environment, that's a restart away rather than something a user can do. Storing the endpoint alongside the verified model and re-testing when it changes would close it.

3. One ask: let a failed capability test degrade rather than refuse. This covers the remaining flexibility question and I think it gets us both what we wanted.

At the moment a failed comparison blocks the save entirely, so a local model that reads a single card perfectly but can't compare two images is unusable. That class of model is common in self-hosted setups, and it was usable before, just without candidate verification.

If a failed test instead saved the configuration with visual verification disabled for that model, you'd keep your "never assume capability" guarantee, since capability would still be proved rather than inferred, and the single-image path would stay open for models that can only do that. It also removes the need for me to argue for a default keyed off openai_requires_key(), which your test supersedes.

Two smaller things in the same area: the comparison is an exact string match, so MAGENTA-GREEN. or a markdown-bolded answer fails despite the model having got it right, and max_attempts=1 means one transient blip fails a save. Both would be less painful if failure degraded rather than refused.

I think that about covers everything I could find!

Git-Romer:
@jasonbayton Thanks for the detailed review. I’ve addressed all points:

• The capability test now uses and validates genuine PNG images rather than checking base64 equality only.
• Capability verification is bound to both the selected model and configured endpoint. Changing either requires a retest, including for already queued scans.
• If multi-image comparison fails but single-image recognition succeeds, an administrator may explicitly accept limited mode using an acknowledgment checkbox. Settings and the card scanner then display a persistent warning, and AI reference-image comparison is disabled.
• Harmless response formatting is accepted, and transient failures use bounded retries.
• Provider, authentication, billing, and model errors cannot be incorrectly treated as limited capability.
• Current main has been merged and the conflicts are resolved.

could you please check again if this fixes all addressed points finalizes this issue?

jasonbayton:
I think that looks good, thank you!

PR #362, opened 2026-08-13
Refuse multi_user_mode on the single-key settings endpoint too

jasonbayton:
Follow-up to #354.

`update_settings` (`PUT /api/settings/`) refuses `multi_user_mode` with a 409 so the setting can only move through `/api/auth/mode`. `backend/api/settings.py` has a second writer, `set_setting` (`POST /api/settings/{key}`), which takes any key, checks admin role, and writes it — so the rule can be walked around:

```
POST /api/settings/multi_user_mode
{"value": "true"}
```

That turns the login screen on or off without the confirmation the settings page shows, and without whatever `/api/auth/mode` does around it. `_apply_setting_side_effect` has no `multi_user_mode` branch either, so nothing compensates downstream.

It is admin-only, so this is not privilege escalation. It is a way for an admin to disable authentication entirely without the warning that #354 was careful to add, which matters most on an install reachable beyond a trusted network.

### The change

Both writers now share one guard, keyed on a small table of settings that belong to a dedicated endpoint, so a third writer cannot quietly miss it:

```python
DEDICATED_ENDPOINT_KEYS = {
    "multi_user_mode": "/api/auth/mode",
}
```

The 409 body is unchanged.

### On the test

`test_generic_settings_endpoint_cannot_change_user_mode` only exercises `update_settings`, so it passes either way and the suite reads as covering this when it does not. The new test calls `set_setting` directly and fails against `main` with `AssertionError: HTTPException not raised`, then passes with the fix.

Full backend suite run against this branch: 585 tests, all passing.


Git-Romer:
Hi Jason, thanks for identifying and fixing this bypass. I’ve updated the branch onto current main while keeping your core fix intact.

Additional changes:

• Both generic settings writers now use the same dedicated-endpoint guard.
• The guard metadata is reusable for future protected settings.
• The regression test now confirms the exact 409 response and that the mode remains unchanged.
• Version bumped to 1.38.4.

Thanks again for catching this.

PR #371, opened 2026-08-23
fix: search the local card catalogue instead of live TCGdex, with fallback

lamiskin:
## What and why

Card recognition's candidate search (`_search_and_rank_candidates` in `backend/api/recognize.py`) called the live TCGdex search API (`GET /v2/{lang}/cards?name=...`) on every scan, even though this app already syncs the full TCGdex catalogue into a local `cards` table that the actual card search page already queries instead. This is redundant, adds latency, and creates an unnecessary external point of failure for something the local DB already answers.

This was found concretely, not theoretically: a scan correctly recognized "Bill" via the vision model, but got zero matches because the container's DNS briefly couldn't reach `api.tcgdex.net` — even though the local catalogue already had the exact matching row (`base4-118`, name "Bill", number 118).

**The fix:**
- Candidate search now queries the local `cards` table first, reusing `accent_insensitive_contains` (`backend/services/text_search.py`) — the same substring-search helper the real search page already uses — rather than any new matching logic.
- Per search pair (the existing energy-substitution / simplified-name / language-fallback logic is untouched — only how each pair is turned into candidates changed): if the local query for that exact `(language, name)` pair returns zero rows, it falls back to a single live TCGdex call for that pair only. This covers two cases that are indistinguishable from the DB alone: the card genuinely doesn't exist, or the local sync just hasn't reached its set yet (a new release, or a full sync still in progress). The fallback result is used for that one scan only and is **not** persisted to the catalogue — the live search response is missing most `Card` columns (types, attacks, abilities, dex_ids, etc.) that a real sync populates, so writing a partial row risked a permanently incomplete entry.
- The fallback is best-effort: a network failure there degrades to no candidates for that one pair, not a broken scan.
- `ScanTrace.record_tcgdex` gained an optional `source` field (`"local"` / `"api_fallback"`) so offline trace analysis can measure how often the fallback actually fires.

### Estimated API calls saved

Each scanned card generates 1–4 search pairs (base language pair, +1 if the simplified name differs from the printed name, +1/+1 for an English-language fallback pair when the detected language isn't English). The search loop doesn't stop after the first pair succeeds — it tries every generated pair until candidates hit a cap of 15 — so before this change, *every* generated pair fired a live call.

For a card whose language is actually synced locally (i.e. in `TCGDEX_SYNC_LANGUAGES`) and whose set has already been synced — the common case — the local query now succeeds on every pair, so the live fallback never fires. That's **1–4 search calls eliminated per card**, essentially 100% of what this step used to spend, for anything already in the catalogue. For a 50-card batch scan, that's roughly **50–200 fewer live search calls** for one batch.

This does *not* touch `_fill_candidate_details` (a separate function, unchanged) which was already local-first before this PR — it only makes per-candidate detail calls for the top few candidates when specific fields (artist/hp/regulation_mark/printed_total) are still missing after checking locally. Nor does it touch `assets.tcgdex.net` image downloads for pHash/visual verification, a separate call path entirely.

## Validation

- [x] Relevant automated checks pass — full backend suite: 698 passed, 9 skipped, 0 failed (`cd backend && pytest`)
- [x] Documentation was updated when behavior changed — `docs/ARCHITECTURE.md` and `docs/BACKEND.md`'s scanner-flow sections now describe the local-first search and the fallback

New test coverage (`SearchAndRankCandidatesLocalDbTests` in `backend/tests/test_recognize.py`) seeds an in-memory SQLite DB rather than mocking HTTP for the local-hit path, and mocks `httpx.AsyncClient` for the fallback path — covering: exact local match, a number match beyond the query cap still floating to the top, English-language fallback, accent-insensitive search, custom-card exclusion, local `Set` metadata attachment, local-hit never calling the live API, local-miss falling back to the live API with correct field mapping, a fallback network failure degrading gracefully, and trace entries correctly tagged by source.

Two pre-existing tests were not mocking `httpx` at all and started making real calls to the live TCGdex API once the fallback existed (one of them returned real "Bill" cards over the actual network mid test run). Fixed to mock an empty response, so they stay deterministic and scoped to what they actually test.

## Card UI (when applicable)

Not applicable — backend-only change, no UI/visual change.


Git-Romer:
Thanks for implementing the local-first candidate search, @lamiskin.

The final implementation uses locally stored catalogue entries whenever their complete normalized printed name matches. It contacts TCGdex only when no compatible local entry exists or when the recognized collector number is missing locally, covering newly released printings before the next catalogue sync.

Substring results from both sources are filtered to prevent incorrect matches such as Mew/Mewtwo or different card suffixes. Collector-number matching also handles padded and alphanumeric formats, crowded result sets, accents, and repeated OCR whitespace.

Existing installations retain catalogue rows with the legacy `is_custom = NULL` state, while actual custom cards and incomplete rows without a TCGdex ID remain excluded. Local database failures now follow the existing rollback and retry path. Live fallback results remain scan-only, with complete card data retrieved when the result is confirmed.

The `v1.41.7` patch bump is included. Thanks for the contribution!

PR #373, opened 2026-08-23
Feature/advanced card search

S-P-A-C-E-B-A-R:
# Add rule text search for cards and collections

## Summary

Adds Rule text search to both Card Search and Collection.

This allows users to search card mechanics across:

- attack names
- attack effects
- ability names
- ability effects
- Trainer and Energy card effects

For example, searches such as `draw 3 cards`, `discard an Energy`, or `once during your turn` can now be used to find cards by what they actually do.

## Why

PokéCollector already stores detailed gameplay metadata for cards, but that information was not searchable.

This supports two related workflows:

- **Card Search:** discover cards with a particular mechanic, including cards a user may want to add to a wishlist.
- **Collection:** find cards with a particular mechanic among cards the current user already owns.

### Search within Collection

Collection Rule text search

### Search within Card Search

Card Search Rule text search

### Backend

- Adds optional `rule_text` filtering to Card Search.
- Adds optional `rule_text` filtering to Collection.
- Searches `card_effect` and the `name` / `effect` fields within `attacks` and `abilities`.
- Shares the JSON-array search implementation between both endpoints.
- Supports PostgreSQL and SQLite.
- Handles SQL `NULL`, JSON `null`, and empty arrays safely.
- Collection results remain scoped to the authenticated user and preserve individual collection rows.

### Frontend

- Adds a Rule text filter to Card Search.
- Adds a Rule text filter to Collection.
- Supports multi-word searches without stripping spaces while typing.
- Collection requests are debounced to avoid querying on every keystroke.
- Existing Collection results remain visible while a Rule text request is refreshed.
- Empty or whitespace-only Rule text is omitted from API requests.

## Testing

Backend coverage includes:

- attack name
- attack effect
- ability name
- ability effect
- card effect
- null/empty JSON handling
- current-user Collection isolation
- preservation of separate Collection rows
- existing Card Search and language behavior

The PostgreSQL-specific JSON query path was also verified against the running Docker PostgreSQL environment using the application's SQLAlchemy query path.

Frontend coverage includes:

- Rule text request parameters
- multi-word Rule text values
- clearing and resetting filters
- Collection query behavior
- Card Search URL-state behavior
- translation-key validation

Before opening this PR I ran:

```text
cd frontend
npm ci
npm test
npm run build


Git-Romer:
Welcome to PokeCollector, @S-P-A-C-E-B-A-R, and thank you for your first contribution! Adding rule text search to Card Search and Collection is a useful improvement for finding cards by their actual mechanics. Your implementation for searching card effects, attacks, and abilities remains intact across both PostgreSQL and SQLite.

I made the following changes on top while preparing the PR:

- Extended rule text filtering to card code and number searches.
- Prevented existing local cards that do not match the rule text from triggering unnecessary TCGdex refreshes.
- Trimmed search input and made `%` and `_` behave as literal characters instead of SQL wildcards.
- Split Card Search into normal and advanced filter sections.
- Improved filter labels, sorting controls, mobile layout, and Collection error handling.
- Made persistent filter panels dynamic across Card Search, Collection, Wishlist, Products, public collections, binder contents, and Sets.
- Added URL synchronization without page reloads or unnecessary browser history entries.
- Preserved unrelated URL parameters and reset pagination when filter changes require it.
- Debounced text-based searches while keeping selections immediate.
- Restored filter state from the URL after refreshes and navigation.
- Kept quick search, sorting, and other lightweight controls immediate.
- Added the `v1.45.0` minor version bump.

Your original commits and authorship are preserved. Thanks again for putting this together. If you know other collectors or developers who might enjoy PokeCollector, spreading the word would help the project a lot!

PR #375, opened 2026-08-25
feat: let custom cards get an owner-uploaded photo

lamiskin:
## What and why

A custom card's only artwork slot (`image_url`) requires a public HTTPS URL, which a physical card with no listing anywhere online can't supply — exactly the case custom cards exist for. `POST /api/collection/{item_id}/photo` refused custom cards outright ("Custom cards already have editable artwork"), and the same assumption was baked into two frontend spots. This lets a custom card get an owner-uploaded photo the same way any other card can.

## What changed

- `POST /api/collection/{item_id}/photo` no longer rejects `is_custom` cards.
- `Collection.jsx`: `hasReferenceArtwork` no longer treats `is_custom` alone as proof of having artwork — `hasApiImage` already covers a custom card that has `image_url` set (stored in `images_small`/`images_large` same as a synced card), so a custom card with none now correctly falls back to a private photo.
- `CollectionCardImage.jsx`'s `showsOwnPhoto` (shared by every grid/list/carousel card view) had the identical `!card.is_custom` exclusion; removed.

## Validation

## Card UI

- [x] I checked `docs/CARD_SYSTEM.md` and followed the established card visual language where it fits — reused the existing "MY CARD PHOTO" upload control as-is, no new UI added
- [x] I included desktop/mobile screenshots or explained why they are not relevant — before/after attached below; no new visual pattern introduced, so no separate mobile shot

## Testing

Backend: new `test_collection_photos.py::CustomCardPhotoTests` confirms a photo can be attached to and read back from a custom card. Full backend suite passes.
Frontend: `CollectionCardImage.test.js` updated with the two custom-card cases for `showsOwnPhoto` (no `image_url` → falls back to the photo; `image_url` set → stays primary). Full frontend suite passes.

## Screenshots
Screenshot_20260825_224536

Screenshot_20260825_224451


Git-Romer:
Thanks, Lachlan. I tested this on desktop and mobile and it works well. I’ve added a few follow-up changes:

- Added browser coverage for uploading, replacing, viewing, switching, and removing photos on manually created cards.
- Kept the pencil as the only collection marker for manually created cards.
- Made sure an uploaded photo follows the card if it is later matched to an official card, without overwriting an existing photo.
- Bumped the version to `1.44.0`.

Your original commit and authorship are preserved. Thanks again — this completes the custom photo feature nicely.

PR #376, opened 2026-08-25
feat: opt-in Gemini fallback when the openai scanner is unconfident

lamiskin:
## What and why

Recognition sometimes comes back without a confident match on the primary OpenAI-compatible provider. Main case this targets: slower consumer-grade hardware running a smaller local model, which reads cards less reliably than a hosted frontier model. This adds an opt-in per-user fallback that retries with Gemini before giving up — a second, independent read turns some of those into a real match, without changing anything for users who haven't turned it on.

## What changed

- `scanner_gemini_fallback` per-user setting (default off). When on, primary provider is OPENAI, and the first attempt comes back unconfident, runs a fresh, independent Gemini extraction + match — never reuses the first attempt's fields, so a weaker read doesn't carry over.
- Non-blocking: a failed fallback attempt (rate limit, expired key, outage) falls back to the original result rather than failing the scan.
- Review UI shows a small "Resolved via Gemini fallback" badge on items where it fired.
- Settings toggle in AI/Card Scanner, next to the existing diagnostics toggle.

## Validation

- [x] Documentation was updated when behavior changed (in-app Settings copy only; no other docs affected)

## Card UI

- [x] I checked `docs/CARD_SYSTEM.md` and followed the established card visual language where it fits — toggle reuses the diagnostics-toggle save pattern, badge matches the existing corner-badge convention
- [x] I included desktop/mobile screenshots or explained why they are not relevant

## Testing

Backend: 8 new tests in `test_recognize.py` (default-off state, happy path, no-key/already-confident skip, already-Gemini no-op, non-blocking failure). Full suite passes.
Verified live against two real Gemini failure modes (503 overload, 429 rate limit), not just mocked.

## Screenshots

image


Git-Romer:
Thanks for adding this, @lamiskin. An optional Gemini second pass is a useful addition for people using OpenAI or a local OpenAI-compatible scanner, especially when difficult photos leave the primary model uncertain.

I kept your core implementation, per-user settings toggle, independent Gemini extraction, and the fallback badge in Scan Review. I made the following changes on top while integrating it with the current scanner pipeline:

- Preserved each provider’s configured model, timeout, capability checks, candidate preloading, and queue caching.
- Made Gemini replace the primary result only when Gemini confidently identifies the card.
- Preserved the original result and candidates when Gemini is also uncertain.
- Made settings, API key, model, rate-limit, parsing, and provider failures fully nonblocking.
- Restored the original queued result if a fallback attempt fails after extraction.
- Kept queued fallback requests at background priority.
- Separated primary and fallback diagnostics so one attempt cannot overwrite the other.
- Added broader regression coverage and the `v1.46.0` minor version bump.

Your original commits and authorship are preserved. Thanks again for continuing to improve the scanner!

PR #377, opened 2026-08-25
fix: raise vision-call timeouts for slow local models

lamiskin:
## What and why

Vision-model calls used a fixed 20-30s timeout inherited from Gemini's latency profile. A local reasoning model — slower consumer-grade hardware, smaller model — can spend 15-45s+ "thinking" through one card before answering (measured up to 120s on a real card with an ambiguous printed symbol), which made recognition and visual verification fail silently on nearly every call for that setup.

## What changed

- New `VISION_REQUEST_TIMEOUT_SECONDS = 180`, applied at all three vision-call sites in `backend/api/recognize.py` (previously 20s/30s/30s).
- Settings' scanner-test call (`backend/api/settings.py`) raised 30s -> 60s — same slow-model concern, smaller test image.
- Scan queue lease (`backend/services/scan_queue.py`) raised 10min -> 20min so a slow call doesn't get reclaimed mid-flight.

## Validation

## Card UI

- [x] I checked `docs/CARD_SYSTEM.md` and followed the established card visual language where it fits — no UI change, backend timeout values only
- [x] I included desktop/mobile screenshots or explained why they are not relevant — not applicable, no visual change

## Testing

687 backend tests pass. No new tests — a timeout duration isn't meaningfully unit-testable without mocking network delay; verified live against real slow-provider calls instead.


Git-Romer:
Thanks for investigating slow local vision models, @lamiskin.

I extended the fix with an advanced AI response timeout setting rather than applying 180 seconds to everyone. It defaults to 30 seconds and offers 60, 120, and 180 seconds, stored separately for each provider.

The selected timeout is now applied consistently across card recognition, batch processing, visual verification, and the connection test. Browser and proxy limits were also aligned so they cannot abandon a valid slow response prematurely.

Backend, frontend, desktop/mobile browser, PostgreSQL, production-build, and smoke tests all pass. The required `v1.41.5` patch bump is included. Thanks for the contribution!

PR #378, opened 2026-08-25
fix: stop vintage/multilingual cards from silently defeating scanner matching

lamiskin:
## What and why

Diagnosed against a real batch of 48 vintage Japanese cards (Neo Genesis-era, ~1999-2001) that all came back unresolved. Four independent, provider-agnostic fixes.

## What changed

**1. Pokédex-number / collector-number confusion.** Older cards, like Example 1 below, print a "No. 0156" National Pokédex reference instead of a fraction-style collector number. The extraction prompt didn't distinguish the two, so the model would read that as `number_local` — which then actively contradicted every real candidate and blocked `_metadata_decision`'s confident-match paths entirely, including `artist_hp`. Prompt now explains the difference explicitly; `normalize_recognized_card_info` also drops any value that still textually reads as a "No." reference as a safety net.

**Example 1** (Pokédex-number confusion):

1_pokedex_number_and_translation_quilava

**2. English-translation fallback ran unconditionally.** The search also queried the model's own English translation of the native name, even when the native-language search already had good results — and that translation can name the wrong species outright. Confirmed on real cards during this same investigation: マグマラシ (Example 1) was independently mistranslated as "Magmar" when it's actually Quilava (No. 156, evolves from Cyndaquil — Magmar is No. 126), and ワンリキー (Example 2) as "Mankey" when it's actually Machop (Mankey's real Japanese name is マンキー). Both read correctly in the native script; the translation step is what got the species wrong. Now only runs as an actual fallback when the native-language search comes back thin, instead of merging bad candidates into results that were already fine. English-language cards are unaffected since `language == "en"` never reaches this branch.

**2b. That same fallback, once triggered, had no cap of its own.** Confirmed on a real card: TCGdex's Japanese catalogue only has two printings of Magmar (マグマラシ — the actual species behind this same name, distinct from the Quilava mislabeling above), so native search came back thin and the fallback filled the *entire* shared 15-candidate budget with 8 translated-English results, burying the 2 genuine native matches in the review grid (both still ranked first, but it visually read as "barely finds the right language"). Added `ENGLISH_FALLBACK_MAX_CANDIDATES` (2x the trigger threshold), enforced on what actually gets *added* — a single search pair's own result limit can already exceed the cap in one call, so checking only between pairs wasn't enough.

**Example 2** (translation fallback naming the wrong species):

2_translation_fallback_machop

**3. No resolution path for Trainer/Energy cards.** `_metadata_decision` could confidently resolve a Pokémon via `artist_hp` (artist + HP match), but Trainer/Energy cards like Example 3 have no HP, so a search that narrowed to exactly one candidate with nothing contradicting it still couldn't resolve. Added a `sole_candidate` path, scoped to only fire after the existing contradiction check already passed.

This one isn't Japanese/vintage-specific — `_metadata_decision` is the single shared matcher used for every scan regardless of language, so an English Trainer/Energy card with no HP was equally unable to resolve via `artist_hp` before this fix. No photo example for that case here: every Wizards-era English card prints a collector number, so a legible product photo would resolve via the existing `number_unique`/`number_metadata` paths on a correct read — `sole_candidate` only fires when that read fails despite a number being printed, which is a property of the OCR attempt, not something a static photo can demonstrate.

**Example 3** (Trainer/Energy card, no HP to match on):

3_trainer_energy_sole_candidate_potion_energy

## Validation

## Card UI

- [x] I checked `docs/CARD_SYSTEM.md` and followed the established card visual language where it fits — no UI change, backend recognition logic only
- [x] I included desktop/mobile screenshots or explained why they are not relevant — not applicable, no UI change; source card photos attached above instead to illustrate the recognition problem itself

## Testing

708 backend tests pass (17 new, covering each fix and its guardrails against regressing English/other-language cards).


Git-Romer:
Follow-up maintainer validation:

- Rebased and integrated the PR with current `main`, preserving the scanner safeguards from #371.
- Tested public real-card photos of Japanese Quilava, Bill, and Double Colorless Energy through the recognition flow.
- Quilava confirmed that printed `No. 156` is correctly treated as a Pokédex reference rather than a collector number.
- The Bill photo exposed an interaction where English fallback candidates prevented the correct unique Japanese Trainer match. Fixed in `6085e34`; replaying the real Gemini extraction now correctly resolves `PMCG1-074`, while multiple native candidates still remain unresolved.
- Double Colorless Energy was extracted correctly but safely remained unresolved because the catalogue lacks an exact entry under its printed Japanese name.
- Full backend suite: 771 passed, 8 expected skips. Version check passes at `1.41.8`, a fresh reviewer found no blockers, and all GitHub checks are green.

Residual limitation: Gemini returned HTTP 503 during the fresh post-fix attempt, so the successful real-photo extraction was replayed through the corrected deterministic matcher. No internet-sourced images were committed.

PR #379, opened 2026-08-25
fix: stop collection add from overriding an explicit card_id language suffix

lamiskin:
## What and why

`CollectionItemCreate.lang` defaults to `"en"`, so when a caller sends a `card_id` that already carries an explicit language suffix (e.g. `"PMCG3-031_ja"`) but omits `lang`, the default silently outranked it. `_add_collection_item` computed `item.lang or detected_lang or "en"` — since `item.lang` defaults to `"en"` rather than `None`, it's always truthy and always wins, so `detected_lang` (parsed from the suffix) never got a chance to matter. The add would target `"PMCG3-031_en"` instead of the requested `"PMCG3-031_ja"`, hit `ensure_card_exists`'s not-found path, and trigger an avoidable live TCGdex fetch (or an outright failure, if the card is genuinely language-specific and has no English printing).

Found while scripting a bulk add of real Japanese cards through the API — every `_ja`-suffixed `card_id` failed until `lang` was passed explicitly alongside it, which shouldn't be required when the suffix already says so.

## What changed

`backend/api/collection.py`, `_add_collection_item`: mirror the precedence `ensure_card_exists` (same file) already uses correctly — an explicit suffix on `card_id` wins over the schema default; the default only applies when `card_id` has no suffix to read.

## Validation

- [x] Documentation was updated when behavior changed (inline comment only; no user-facing docs affected)

## Card UI

- [x] I checked `docs/CARD_SYSTEM.md` and followed the established card visual language where it fits — not applicable, backend request-handling only
- [x] I included desktop/mobile screenshots or explained why they are not relevant — not applicable, no UI change

## Testing

New regression test `test_add_collection_item_prefers_composite_id_suffix_over_default_lang` (confirmed it fails without the fix, passes with it). Full backend suite (707 tests) passes.


Git-Romer:
## What and why

`CollectionItemCreate.lang` defaults to `"en"`, which caused an explicit language suffix such as `_ja` to be overwritten when adding a card. This could select the wrong card, trigger an unnecessary TCGdex request, or make the addition fail.

The original patch corrected the CSV write helper, but the normal and bulk collection endpoints contained the same duplicated precedence logic.

## What changed

- Added one shared language resolver for collection additions.
- Applied it to single-card, bulk, and CSV additions.
- A recognized language suffix in `card_id` is authoritative.
- Unsuffixed IDs continue using and validating the requested language, defaulting to English.
- Custom cards continue deriving their language from their database record.

## Safety and coverage

Coverage includes every supported language suffix, hyphenated languages, aliases, defaults, conflicting values, invalid values, and all three collection-add paths.

## Release

Includes the required patch release metadata for `v1.41.3`.

Git-Romer:
Thanks for catching this while bulk-adding Japanese cards, Lachlan. Your regression case was spot on. During review I found the same language-precedence problem in the normal and bulk endpoints, so I expanded the shared fix and coverage across all three collection-add paths.

Everything is green now, including the production smoke tests, and the fix is merged as v1.41.3. Thanks for the contribution!

PR #380, opened 2026-08-25
fix: cap concurrent scan processing and raise the upload size limit

lamiskin:
## What and why

Two independent operational fixes found while load-testing a real 48-item batch scan.

**1. Unbounded concurrent scan processing exhausted the DB pool.** `process_claimed_scan_item()` holds a DB session open for the entire recognition call, including the vision-model HTTP round trip. Each retry request schedules its own independent background task, and FastAPI runs those concurrently with no cap of its own — a "retry all" click, or several jobs finishing their backoff around the same time, could each hold a connection open for minutes at once, especially against a slower provider. Measured hitting real `QueuePool limit ... overflow 20 reached` at a few dozen concurrent retries.

**2. Upload size limit mismatch caused silent 413s.** The `/api/` nginx location's `client_max_body_size` (100M) was lower than the server-level default (500M) it should have matched, so a full-size batch upload at realistic phone-camera resolutions got rejected before it ever reached the backend.

## What changed

- `backend/services/scan_queue.py`: `MAX_CONCURRENT_SCAN_PROCESSING = 3` via `asyncio.Semaphore`, wrapping the DB-session-holding portion of `process_claimed_scan_item()`. Capped at 3, not tuned higher — a single-instance local provider processes one request at a time regardless of how many arrive together, so allowing more concurrency wouldn't add throughput, only cost pool headroom.
- `frontend/nginx.conf`: `/api/` location's `client_max_body_size` raised 100M -> 500M to match the server-level default.

## Validation

## Card UI

- [x] I checked `docs/CARD_SYSTEM.md` and followed the established card visual language where it fits — not applicable, backend/infra only
- [x] I included desktop/mobile screenshots or explained why they are not relevant — not applicable, no UI change

## Testing

Both are infrastructure/reliability fixes not meaningfully unit-testable (a semaphore bound, an nginx config value) — verified by reproducing both failures against a real batch scan and confirming they stopped.


Git-Romer:
Thanks for investigating and reproducing this with a real 48-card batch, @lamiskin.

During validation I found that the concurrency limiter also needs to work across FastAPI and APScheduler’s separate event loops. I adjusted it to maintain an exact process-wide limit of three and added coverage for load, cancellation, failures, database-session cleanup, and scheduler/web overlap.

I also scoped the larger nginx allowance specifically to the batch-upload endpoint. It now permits 205 MiB—enough for the backend’s 200 MiB batch limit plus multipart overhead—while other API routes remain limited to 100 MiB.

All backend and frontend tests, PostgreSQL checks, production builds, nginx limit probes, and smoke tests pass. The required `v1.41.4` patch bump is included. Thanks for tracking down both failures!

PR #382, opened 2026-08-25
fix: never composite photos for a provider that can't read multiple images

lamiskin:
## What and why

A composite grid asks the model to read several cards out of one image, which needs the same multi-image capability visual verification already depends on (`scanner_capability_mode`, proven by the `/settings/scanner/test` probe). A provider proven only single-image-capable (`degraded` mode) can't do this reliably either.

`recognize.py`'s single-photo path already respects `capability_mode`, but the batch enqueue path in `scan_jobs.py` called `require_scanner_capability_mode(...)` only for its validation side effect and discarded the returned mode, so a degraded provider could still be asked to composite whenever the client's per-photo `individual` toggle wasn't set.

This makes the batch path use the mode it was already fetching:

```python
capability_mode = require_scanner_capability_mode(
    db, current_user.id, provider.name, provider.model()
)
...
batch_modes = [
    len(files) > 1
    and position not in individual_set
    and capability_mode != SCANNER_CAPABILITY_DEGRADED
    for position in range(len(files))
]
```

This is defense in depth — the server no longer just trusts the frontend's own toggle state. The frontend also hides the now-pointless "process individually" toggles (per-photo and toggle-all) when the configured provider is in degraded mode, since the server already forces individual processing there and the toggle would have nothing left to override.

Found while auditing an abandoned pre-#355 branch: most of its Gemini-specific retry/UI text was already superseded by #355's provider-neutral rewrite, but this compositing-vs-capability gate was never carried over.

## Validation

- [x] Added `test_degraded_capability_forces_every_photo_individual`, asserting the server forces every photo through individually for a degraded provider even when the client claims none are individual
- [ ] Documentation was updated when behavior changed (no user-facing docs affected)

## Card UI (when applicable)

- [x] Not applicable — this only changes which existing toggles are shown/hidden, no new visual pattern


Git-Romer:
Thanks for finding the missing capability gate in batch scanning, @lamiskin.

I validated the fix against current `main` and extended it to jobs that were already queued before a provider was changed or retested. Verified limited providers now remain individual both when queued and immediately before processing, without sending an accidental composite request. Unverified configurations remain blocked until Test and save succeeds.

I also added desktop and mobile coverage for full, limited, and unverified states and fixed the floating navigation button so it cannot cover the scanner action on mobile.

All backend, frontend, browser, PostgreSQL, production-build, and smoke tests pass. The required `v1.41.6` patch bump is included. Thanks for the contribution!

PR #384, opened 2026-08-26
fix: paint an opaque canvas so Firefox on Android has no white bars

lamiskin:
## What and why

The app had no opaque `background-color` on `html` or `body` — the dark appearance depended entirely on a gradient image. `body`'s winning background rule uses the unlayered `background` shorthand, which resets `background-color` to transparent, so the canvas behind that gradient was the browser's default white. Chromium extends the gradient across overscroll/safe-area regions; Firefox on Android doesn't, showing white bars.

Sets `background-color: var(--color-bg)` on `html` (whose background always paints the whole canvas, themed per `[data-theme]`) and restores an opaque fallback colour to `body`'s own shorthand.

Also found on-device: a persistent white strip on the *right* edge, at rest rather than only during a bounce gesture — a different bug, since nothing at the page root should ever scroll horizontally. Painting over it would have hidden a real overflow bug rather than fixed it, so this adds `overflow-x: hidden` to `html, body` to remove that scroll surface entirely.

**Before** — white strip visible on the right edge:
After

**After** — fixed:
Before

## Validation

- [x] Confirmed on-device (Firefox on Android) before and after
- [ ] Documentation was updated when behavior changed (no user-facing docs affected)

## Card UI (when applicable)

- [x] Not applicable — global canvas background fix, no new visual pattern


Git-Romer:
Thanks for tracking this down and confirming it directly in Firefox on Android. Your explanation of the transparent root canvas and the separate horizontal overflow case made the cause very clear.

I brought the branch up to date with the current main branch while keeping both of your fixes intact. The root and body now always paint an opaque themed background, and the page root cannot pan horizontally. I also added the required patch version bump to 1.42.2.

Thanks again for taking the time to test this on the affected browser and document both symptoms so clearly.

PR #385, opened 2026-08-26
feat: zoomable card artwork in the detail dialog

lamiskin:
## What and why

The card detail dialog only ever showed the artwork at its fixed sidebar size — no way to look closer at a scan/border/print detail. Adds a full-screen pan/zoom viewer for that image: scroll wheel or pinch to zoom, drag to pan once zoomed, a single click/tap on the image toggles zoom in/out, and clicking outside the image closes the viewer. On-screen +/-/reset controls double as the visual hint that the image is interactive, alongside a small magnifying-glass badge on the thumbnail itself and a `cursor-zoom-in` affordance.

Single click (not double-click) was a deliberate choice to match the existing single-click "enlarge" pattern already used by the scan candidate compare view (`ScanZoomModal`/`Maximize2` in `ScanReview.jsx`), for consistency across the app rather than introducing a second convention.

New `ImageZoomOverlay` component, wired into `UnifiedCardDialog` (`components/UnifiedCard.jsx`) so every consumer of the shared detail dialog gets it for free. Two implementation notes worth a reviewer's attention:
- The overlay is a separate `createPortal` from the dialog itself (both target `document.body`), so it's a DOM *sibling*, not a descendant, of the dialog. The dialog's own document-level Escape listener doesn't know about the overlay, so closing it also fired the outer dialog's close — fixed by having the dialog track (via a ref, so the listener closure stays current) whether the overlay is open and skip its own close while it is. First Escape closes the zoom, second closes the dialog.
- The `` itself is `pointer-events: none` (needed so drag/wheel/pinch handlers on the pan surface behind it keep working), so "clicked the image" vs. "clicked the backdrop around it" is decided by comparing the click point against the image's actual rendered bounding rect rather than `event.target`.

## Validation

- [x] Relevant automated checks pass — full frontend suite (`npm test`) and `npm run build` both pass. No dedicated interactive tests were added for the new pan/zoom/click-toggle behavior itself: this repo has no `@testing-library/react` (or similar) dependency, and the existing `UnifiedCard.test.js` only exercises pure functions and `renderToStaticMarkup` output for that reason — `UnifiedCardDialog` itself (which uses `createPortal`) isn't tested there either, so this follows existing convention rather than introducing a new one. Verified manually end-to-end instead: opened the dialog, clicked the zoom badge, confirmed single-click zoom in/out, click-outside-close, click-on-image-doesn't-close, drag-to-pan not misfiring a close, and the two-stage Escape behavior — all against a real running instance.
- [ ] Documentation was updated when behavior changed — no user-facing docs affected.

## Card UI (when applicable)

- [x] I checked `docs/CARD_SYSTEM.md` and followed the established card visual language where it fits — reused the existing dark-pill icon-button treatment (`border-white/15 bg-black/75`) already used for the scan-compare "Maximize2" badge, so the new controls read as the same family rather than a new visual idea.
- [x] I included desktop/mobile screenshots or explained why they are not relevant 
- [x] If this introduces a genuinely new visual idea, I explained why the existing patterns do not fit — this is a new feature-specific interaction (full-screen pan/zoom), not a new `CardDisplay` variant, so it isn't added to `CardSystemGallery.jsx`; it layers onto the existing shared `UnifiedCardDialog` rather than replacing or forking it.

Screenshot_20260826_224900

Screenshot_20260826_224820


Git-Romer:
Thanks for this, Lachlan. I tested the zoom on desktop and mobile and made a few follow-up changes:

• Kept the normal cursor outside the image and limited the zoom and grab cursors to the artwork.
• Added a dragging cursor and a small movement threshold so clicks and taps are not lost.
• Improved how the image fits on short landscape screens.
• Kept the existing image retry flow working and only enabled zoom once the artwork has loaded.
• Improved the modal focus and accessibility behavior.
• Added browser tests for the zoom controls and gestures.
• Bumped the version to 1.43.0.

Your original commits and authorship are preserved. Thanks again, this is a really nice improvement to the card detail view.

PR #387, opened 2026-08-29
Report a catalogue outage instead of reporting no matches

### LINKED ISSUE #389: A TCGdex outage is reported to the user as 'no matches found'
### What happens

While `api.tcgdex.net` is unreachable, every scan completes as `done` with **zero matches and no error**.

From the user's side that is indistinguishable from "this card is not in the database", so the natural conclusion is that your own installation is broken. That is where the time goes: I checked the scanner provider, the API key, the queue, and the local database before thinking to check whether the catalogue itself was up.

### Evidence

During the outage on my install, a scan of Sandshrew 027/165 produced:

- `recognized` populated correctly — the vision model read the card fine
- `matches` empty, `error` empty, `status` = `done`
- the card present in the local database as `sv03.5-027_en`

So recognition worked, the data existed locally, and the user was still told nothing was found.

### Why it happens

`_search_and_rank_candidates` treats an unreachable catalogue as an empty result:

```python
cards = response.json() if response.status_code == 200 else []
...
except Exception as exc:
    ...
    continue
```

A transport timeout, a 5xx, and a genuine empty result all collapse into the same silent outcome.

### What I would expect

A scan that could not be matched **because the catalogue was unreachable** should say so, and ideally be retried rather than permanently recorded as unmatched — an outage is temporary, and the queue already has transient-error handling for exactly this.

The distinction that matters is between "the catalogue says this card does not exist" and "we could not ask the catalogue". Only the first is a real answer.

I have a patch in #387.

jasonbayton:
Fixes #389.

### The user problem

While `api.tcgdex.net` was unreachable, every scan completed with **zero matches and no error**. From the user's side that is indistinguishable from "this card is not in the database", so the natural conclusion is that your own installation is broken. I spent a while looking at a perfectly healthy install before checking the catalogue.

Candidate search treats an unreachable catalogue as an empty result:

```python
cards = response.json() if response.status_code == 200 else []
...
except Exception as exc:
    ...
    continue
```

A timeout, a 500, and a genuine empty result all collapse into the same silent outcome.

### The change

When **every** lookup failed to reach the catalogue and nothing was found, raise 503.

`_scan_error_from_http` already classes 503 as a `TransientScanError`, so the item is retried with backoff and recovers on its own once the catalogue is reachable, rather than being permanently recorded as unmatched. That felt better than a permanent failure: an outage is temporary, and the scan should survive it.

Two deliberate limits so this cannot become a retry loop:

- A **genuine empty result still reports no matches**. The catalogue answered; the card simply is not there.
- A **4xx is treated as an answer, not an outage**. Retrying a rejected request forever would not help, so only 5xx and transport errors count as unreachable.

### Tests

Four tests: transport failure, catalogue 5xx, and two bystanders covering the genuine-empty and 4xx cases so an over-broad fix fails.

They fail against `main` and pass with the change.

### One note on the suite

`test_own_photo_priority.test_equivalent_prints_flags_the_owned_photographed_variant` currently errors on `main`, with and without this change, and takes ~60s because it makes real calls to `api.tcgdex.net`. That looks like a pre-existing dependency on the live catalogue rather than anything to do with this PR, but it is worth knowing that a chunk of the suite goes red whenever TCGdex is down. Happy to look at that separately if useful.


jasonbayton:
Pushed two corrections after a further review pass.

**A lookup that failed after its response arrived was counted twice.** A gateway returning an unreadable body with a 200 scored two attempts and one failure, so the counts never matched, the 503 was not raised, and the scan completed with no matches. That is the original bug surviving in exactly the case where the catalogue is broken rather than absent. The attempt is now counted once, where it begins.

**A 429 was treated as a genuine empty answer.** The catalogue declining to answer a request is not a statement that the card does not exist, so a rate-limited lookup reproduced the same misleading outcome this PR exists to remove. 429 and 408 now count as unreachable; any other 4xx is still an answer, which the existing test continues to cover.

Both are covered by tests confirmed to fail against the previous code and pass against this one. Full backend suite green (701 tests).

jasonbayton:
One consequence of this PR worth flagging before merge: a catalogue outage now fails *after* the vision extraction has already been paid for, and the queue retries transient failures indefinitely until the fourteen-day expiry. That works out at roughly sixty extractions of the same photo during a sustained outage, none of which can succeed, where before this PR it was one.

Raised as #391 with a proposed fix (bound that reason alone, in its own counter, at four attempts) rather than folding it in here, since it touches the queue and adds a column and you may prefer a different approach. I have it running in my fork with tests if you want it as a PR on top of this one.

Git-Romer:
Thanks for the careful report, and especially for following up with #391. You were right that simply turning the catalogue failure into a retry would cause the same paid vision request to run again.

I updated the implementation on current main and kept the important distinction from your original fix. Genuine empty results and normal 4xx responses still mean no match, while connection failures, invalid responses, 408, 429, and server errors are treated as temporary catalogue outages.

The queue now saves the completed recognition result before catalogue matching. If TCGdex is unavailable, later attempts reuse that result and retry only the matching step. This also works for batch scans, so the same photo is not sent to the paid vision provider repeatedly.

I also tightened the queue ownership and recovery handling so interrupted or expired workers cannot reuse or overwrite stale recognition data.

This addresses #389 and #391. Issue #390 was already resolved by the local-first catalogue matching added in #371.

Thank you for catching both the misleading no-match behavior and the retry cost. Your Sandshrew example and the detailed retry analysis were very helpful in getting this right.

PR #388, opened 2026-08-29
Fall back to the synced catalogue when the remote one is unreachable

### LINKED ISSUE #390: Scanning fails during a TCGdex outage even though the catalogue is synced locally
### What happens

Scanning stops working entirely when `api.tcgdex.net` is unreachable, even on an installation that has already synced the catalogue and holds the card data locally.

### Evidence

On my install during the recent outage:

- local `cards` table: **23,671 cards**, 222 sets
- the scanned card, Sandshrew 027/165, present locally as `sv03.5-027_en`
- scan result: zero candidates

The data needed to match that scan was in Postgres, milliseconds away, and unused. Candidate search in `_search_and_rank_candidates` only ever queries `api.tcgdex.net`, so a remote outage takes scanning down with it despite a complete local copy.

### Why it is worth addressing

The sync exists precisely so the installation holds its own copy of the catalogue. Depending on the remote API for every scan makes that copy decorative for this purpose, and makes a single external host a hard dependency for a core feature.

It also affects anyone running in a restricted network, not only during an outage.

### What I would expect

When the remote catalogue cannot be reached, fall back to the synced copy rather than reporting nothing.

Two limits seem important:

- **A live answer should not be second-guessed.** If the catalogue replied and had nothing, that is a real "no match"; substituting a local guess would hide genuine gaps.
- **Custom cards should be excluded**, since they are the installation's own rows rather than catalogue entries, and offering one as a scan match would invent a result the live search could never produce.

Related to #389, which covers the reporting side of the same outage. I have a patch in #388.

jasonbayton:
Fixes #390.

Builds on #387 — please merge that first, as both touch `_search_and_rank_candidates` and this branch contains its commit.

### The user problem

An installation that has synced the catalogue already holds the card data locally. Candidate search only ever queried `api.tcgdex.net`, so when that host was unreachable, scans reported nothing even though the exact card was sitting in the local `cards` table.

Concretely, during the recent outage: a scan read *Sandshrew 027/165* correctly, `recognized` was populated, and the local database contained `sv03.5-027_en` — and the result was still zero matches.

### The change

When every remote lookup fails to reach the catalogue, the synced copy is searched instead, returning candidates in the same shape so ranking and downstream handling are unchanged. Images come from the synced rows, so they are exactly as available as they were before.

Two deliberate limits:

- **A live answer is never second-guessed.** If the catalogue replied and had nothing, that is still no match. Substituting a local guess there would hide genuine gaps in the catalogue.
- **Custom cards are excluded.** They are the installation's own rows rather than catalogue entries, and offering one as a scan match would invent a result the live search could never produce.

The 503 from #387 still applies when the remote catalogue is unreachable *and* the local one has nothing to offer either.

### Tests

Five tests, including two bystanders: that a live empty answer is not replaced by a local guess, and that the candidate shape matches the live search so ranking cannot silently misread it. The fallback test fails when the fallback is disabled.

### Note on the suite

`test_own_photo_priority.test_equivalent_prints_flags_the_owned_photographed_variant` errors on `main` with or without this change, and takes ~60s because it calls `api.tcgdex.net` for real. Same pre-existing issue I mentioned in #387.

### DIFF
--- a/backend/api/recognize.py
+++ b/backend/api/recognize.py
@@ -32,7 +32,7 @@ from services.scan_providers import (
 )
 import logging
 from fastapi import APIRouter, UploadFile, File, HTTPException, Depends
-from sqlalchemy import or_
+from sqlalchemy import and_, or_
 from sqlalchemy.orm import Session
 from api.auth import get_current_user
 from database import get_db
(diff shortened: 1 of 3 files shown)


jasonbayton:
Rebased onto the corrected #387 and pushed four fixes from a further review pass. Three of them could have filed a **wrong** card rather than merely missing one, so they are worth reading before this is merged.

**Truncation before narrowing.** The rows were cut by an unordered `LIMIT 60` before the printed collector number was considered, so the scanned printing could be dropped while a different card of the same name survived. A single surviving result is exactly what the ranker reads as a unique number and marks confident, and "add all confident" then files it. The rows are now ordered so the set is deterministic, and the number is applied before the cut, using `normalize_scanner_card_number` so the fallback and the live search cannot drift apart on what counts as the same number.

When the number is known and no local row carries it, the fallback now offers **nothing**. The live search can safely return same-name results because it is searching the whole catalogue; this is a partial local copy, so a same-name row of the wrong number is a wrong answer here.

**The fallback ran on any unreachable lookup.** A catalogue that answered "no such card" in one language and timed out in another was overruled by a local guess. It now runs only when every lookup failed to reach the catalogue, matching the condition the 503 already used.

**The (language, name) pairs were searched as their product.** A name only ever searched in English could match a row in another language. There is a real collision in the published catalogue: `swsh2-201` is Milo in English and Yarrow in Italian, while `swsh7-201` is Milo in Italian, all numbered 201. An Italian scan of Yarrow admitted the unrelated Italian Milo. The pairs are now searched as pairs.

**`is_custom` is nullable.** Requiring `IS FALSE` excluded NULL rows, which the card-id migration in `database.py` already treats as catalogue data, so rows predating the flag were invisible to the fallback.

The tests now run against a real SQLite session rather than a mocked one, because the custom-card test previously asserted only that `filter()` had been called, which any filter satisfies. Each fix was mutation-tested: removing it fails its own test and leaves the others passing. Full backend suite green (714 tests).

Git-Romer:
Thanks for putting this together and for the follow-up review that caught the cases which could select the wrong card.

I compared each behavior in this branch with current main. #371 now searches the synced catalogue first, so locally available cards remain matchable when TCGdex is unavailable. It also covers collector-number normalization, language and name pairing, custom-card exclusion, and mixed catalogue-response handling.

One intentional difference remains. A same-name local candidate with a conflicting number may be shown for manual review, but the contradiction prevents it from being marked confident or added automatically. This keeps the result useful without risking a wrong automatic match.

The remaining outage reporting and repeated paid-recognition problem is now handled by #387. There is no separate change left in this branch that still needs to be merged, so I am closing it as superseded by #371 and #387.

Thank you for documenting the Sandshrew case and carefully reviewing the fallback behavior. Those findings directly improved the final implementation.

PR #405, opened 2026-09-13
fix: restore silently reports success after a full backup, due to an orphaned image_cache sequence

ninjasockmonkey:
## What and why

`GET /api/backup/download` excludes `image_cache` from a "full" backup via `--exclude-table image_cache`. `pg_dump` doesn't treat a table's owned sequence as excluded along with it, so the `--clean` output ends up with an orphaned `DROP SEQUENCE public.image_cache_id_seq;` with no matching `DROP TABLE public.image_cache;` to go with it.

On restore, that orphaned `DROP SEQUENCE` fails (the live table's `id` column default still depends on the sequence), and because it runs early in the script, every statement after it - all of the actual collection/user/card data - never executes. `POST /api/backup/restore` didn't catch this either, because its `psql -f` call has no `-v ON_ERROR_STOP=1`, so `psql` just logs the error, keeps going, and still exits 0 - the endpoint reports "Database restored successfully" while the database is untouched.

This PR:
- Switches `--exclude-table` to `--exclude-table-data` for `image_cache`, which still skips the (potentially large) cached image blobs but keeps the table/sequence DROP+CREATE statements correctly paired.
- Adds `-v ON_ERROR_STOP=1` to the restore's `psql` invocation, so a genuinely broken dump surfaces as a real error instead of a false success.

## Validation

- [x] Reproduced against a live install: downloaded a full backup, restored it, got "Database restored successfully" with no data actually restored. Ran the same dump manually with `-v ON_ERROR_STOP=1` and confirmed the exact failure (`cannot drop sequence public.image_cache_id_seq because other objects depend on it`).
- [x] Stripped the orphaned sequence statements from the same dump and confirmed the restore then completes cleanly.
- [ ] No automated backend test suite is wired into CI for this repo, so no existing check to run against this change.


Git-Romer:
Hi @ninjasockmonkey 👋

Thank you for tracking this down and providing the fix! The orphaned `image_cache` sequence was a real issue, and excluding only the table data is the correct solution because it keeps the table and its sequence intact while still avoiding the large cached image payload.

I kept your two core changes and added a few follow-up improvements while integrating the PR with the current version of PokeCollector:

- Restore operations now run inside a single transaction and stop on the first SQL error. A failed restore therefore leaves the existing database unchanged instead of partially applying the dump.
- Restore uploads are streamed into private, uniquely named temporary files rather than being loaded entirely into memory.
- Temporary restore files are removed after successful restores as well as errors, timeouts, interrupted uploads, and missing PostgreSQL tools.
- PostgreSQL connection URL parsing is now shared with the existing pre-upgrade backup system and correctly handles encoded credentials, driver-qualified URLs, and IPv6 hosts.
- The branch was brought up to date with the current Deck Builder release and prepared as patch version `v1.47.1`.

Your original commit and authorship are preserved. Thanks again for investigating the restore behavior and contributing the core fix! 🙌

